PR #5, opened 2017-08-04
Fix tiny typos.

LucyMcGowan:
* the easiest it will be 👉 the _easier_ it will be
* it might be multiple months before your initial report and when we start working on it 👉 it might be multiple months _between_ your initial report and when we start working on it
* Make sure you use tag your tweet 👉 Make sure you tag your tweet
* you can then wrangle and visualise using the tools you already know for. 👉 you can then wrangle and visualise using the tools you already know.

Additionally at the bottom of contribute.md under the "Contribute code" section there is an unfinished sentence.
> We don't want you to  

but I'm not sure what it's meant to say, so I've left that for now.

I haven't re-rendered anything yet, just updated the markdown.

### DIFF
--- a/content/contribute.md
+++ b/content/contribute.md
@@ -14,9 +14,9 @@ If you're interested in answering questions, two good places to start are the ti
 
 ## File issues {#issues}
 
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!  No need to re-render anything - it all happens automatically 😁

PR #7, opened 2017-08-07
Fix a few more small typos

LucyMcGowan:
@hadley I gave another quick read through and found a few more typos:
* can be tremendously _encourage_ 👉 can be tremendously _encouraging_
* You might answer the question, but you'll help someone answer it more easily. 👉 You might _not_ answer the question, but you'll help someone answer it more easily.
* to use in _the every_ data analyses 👉 to use in _the every day_ data analyses (**I'm just guessing this is what you meant, so you may want to double check this sentence**)
* There are a handful _for_ other packages 👉 There are a handful _of_ other packages

I also noticed this small grammatical issue in the packages.md file:
> tibble is a modern re-imaginging of the data frame, keeping what time **has** proven to be effective, and throwing out what **is** not.

I'm not sure if you really care about this change of tense (from has to is) I think it would be proper to state "keeping what time **has** proven to be effective, and throwing out what it **has** not" but that kind of sounds weird 🤷‍♀️. Also probably most people don't care about this level of grammar. ALSO it's possible I'm misremembering my grammar rules. 

Big 🙌 for netlify automagically rendering - I need to switch to that for my casual blogging!

### DIFF
--- a/content/contribute.md
+++ b/content/contribute.md
@@ -6,9 +6,9 @@ The tidyverse would not be possible without the contributions of the R community
 
 ## Answer questions {#answers}
 
-The easiest way to help out is to answer questions. You won't know the answer to everything, but that's ok! Even just the acknowledgement that someone cares enough to try can be tremendously encourage.
+The easiest way to help out is to answer questions. You won't know the answer to everything, but that's ok! Even just the acknowledgement that someone cares enough to try can be tremendously encouraging.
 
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!  

Btw you don't need to mention each individual change since I'll see that in the commits. 

Do you want to fix the tibble grammar issue too? As far as I know, your analysis is correct.

(Also I think I meant "use in every day data analyses")

PR #26, opened 2017-08-15
Use the actual year this was written

jimhester:
I was apparently being optimistic when I wrote the date.

I guess this might break links to the article, not sure there is much we can do about it...


hadley:
Can you try setting up a redirect? https://www.netlify.com/docs/redirects/

jimhester:
Closing because I inadvertantly included this commit in https://github.com/tidyverse/tidyverse.org/commit/44fe0dc1852723cdf7de746946be3426ad5d042a. Murphy's law has a 100% probability when I try to use git GUIs.

PR #39, opened 2017-08-31
add drob's datacamp case study course

robinsones:
Adding David Robinson's datacamp course. Is tidytext officially part of the tidyverse yet? If so, I'll add their book to the book list.


hadley:
Thanks! 

Tidytext is tidyverse adjacent. Maybe a separate category?

robinsones:
That makes sense! Are there any other tidyverse adjacent packages that I could put with it?

hadley:
Not off the top of my head

PR #42, opened 2017-09-07
Show magick code to produce images needed for articles

jennybc:


hadley:
Oh funny, I've just been using preview!

PR #45, opened 2017-09-20
Updated the events post-it

carneybill:
Deleted the classes this week which makes rstudio::conf show up.

### DIFF
--- a/data/events.yaml
+++ b/data/events.yaml
@@ -7,20 +7,10 @@ events:
   what: |
     If you see this event, something has gone wrong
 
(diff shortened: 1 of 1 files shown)


hadley:
Hmmmm they're supposed to disappear automatically. Let me see what's up with that code.

hadley:
Ooooh, they do but the site just needs a manual regen.

hadley:
So there's basically no reason to keep the old events around and we might as well proceed by removing manually, which is a bit of a bummer.

Issue #36, opened 2017-08-29
The package icon is missing for the article categories


LucyMcGowan:
I believe this is because the category specified in the `css` is "package", whereas the category specified in the articles is "packages":
https://github.com/tidyverse/tidyverse.org/blob/21f82db97fac49163c8d4b5307b557f00bdbaf64/static/css/tidyverse-site.css#L802-L803

I suppose it'd be easier just to change the css? I'm happy to do the 📦 PR, just wanted to run this by you first.

hadley:
I'd prefer to avoid changing the css since it's generated by Robby and I just include a copy.

LucyMcGowan:
ah 👌 then should everyone change to "package" instead of "packages" for the category?

I think this effects 3 articles:
* http://www.tidyverse.org/articles/2017/08/googledrive-initial-release/  
* http://www.tidyverse.org/articles/2017/08/tidyr-0.7.0/  
* http://www.tidyverse.org/articles/2017/08/purrr-0.2.3/

I just felt like it looks a bit sad empty :confused:
<img width="243" alt="no_pkg_icon" src="https://user-images.githubusercontent.com/8431897/29840608-34984b68-8cc8-11e7-9a23-d78061b9f107.png">

hadley:
Yes please!

PR #49, opened 2017-10-31
Lubridate 1.7.0

vspinu:
Also fixing [package] category in tidyr erratum as per #36.


content/articles/2017-11-01-lubridate-1.7.0.Rmd
```diff
@@ -0,0 +1,135 @@
...
+library("lubridate")
+knitr::opts_chunk$set(collapse = TRUE, comment = "##=>")
+```
+
+We are pleased to announce that `lubridate` 1.7.0 is now on CRAN! For a complete
+set of changes please see the release
+[NEWS](https://github.com/tidyverse/lubridate/releases/tag/v1.7.0). Here is a
+brief walk-through of the prominent new features.
+
+## Built-in CCTZ and much faster `update` and `force_tz`
```
hadley: Does this mean lubridate now has a C++11 dependency?

content/articles/2017-11-01-lubridate-1.7.0.Rmd
```diff
@@ -0,0 +1,135 @@
...
+
+By default `cutoff_2000` is 68 to comply with R's `strptime` function.
+
+This feature was not propagated to higher level parsing function because base
+`strptime`, on which those occasionally rely, doesn't support this option.
+
+
+## Functions `wday` and `month` are now localized
+
+When `label=TRUE` `wday` and `month` extractors return localized strings:
```
hadley: This is unfortunately going in the opposite direction to the rest of the tidyverse - I think this is dangerous because it makes it easy to write code that works for you, but not in a different locale.

content/articles/2017-11-01-lubridate-1.7.0.Rmd
```diff
@@ -0,0 +1,135 @@
...
+wday(x, label = TRUE, abbr = FALSE)
+wday(x)
+wday(x,  week_start = 2)
+```
+
+## Rounding to fraction of a second
+
+R's date-time format (POSIXct) is accurate up to microseconds, thus it makes
+sense to round to fractions of a second. Unfortunately R currently prints
+fractional seconds incorrectly which can lead to confusion:
```
hadley: Maybe say "truncates" instead of "rounds" rather than incorrect?

content/articles/2017-11-01-lubridate-1.7.0.Rmd
vspinu: Yes. Rcpp dependency with C++11 standard. Are there any issues with that?

content/articles/2017-11-01-lubridate-1.7.0.Rmd
vspinu: I agree. I am avoiding localization as much as I can but in this case it's pretty harmless  as human readable output should never be used programaticaly, only for printing and reporting. Programs should use numeric output which is the default.  

I actually was planing to add a comment on this in the blog, but removed it eventually.

content/articles/2017-11-01-lubridate-1.7.0.Rmd
vspinu: It's a printing issue. I don't know the exact nature of the bug but I don't think it's as simple as truncation. As this example ilustrates R prints ".02" instead of ".03", so it's really an incorrect printing of decimals. No time to dig it further but I have reported it [here](https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17350).

content/articles/2017-11-01-lubridate-1.7.0.Rmd
hadley: There's a minor issue because some risk averse organisations use RHEL, and one of the common stable versions has a very old version of gcc. I think we can just wait to see if this is a problem.

content/articles/2017-11-01-lubridate-1.7.0.Rmd
hadley: Yeah, I think you should add a comment here - i.e. you should only use these for labels, not for comparisons, otherwise your code might not work on someone else's computer.

content/articles/2017-11-01-lubridate-1.7.0.Rmd
hadley: `format.POSIXct` has:

> Specific to R is %OSn, which for output gives the seconds truncated to 0 <= n <= 6 decimal places (and if %OS is not followed by a digit, it uses the setting of getOption("digits.secs"), or if that is unset, n = 0).

hadley: So it's technically not a bug, but documented behaviour 🙄 which means R-core is unlikely to fix it. I suspect they would accept a patch though, and I suspect it's probably not a lot of work.

vspinu: That doc just states that it truncates to digits.secs. In my example digits.secs is set to 6, so it should print .03 as .03 not as .02. It's really a bug.

hadley: Well there's a bug that it doesn't respect `digits.secs`, but the truncation is by design. 

(This is all buggy behaviour but describing it as a bug to R-core when it's (semi) documented will trigger them)

content/articles/2017-11-01-lubridate-1.7.0.Rmd
vspinu: Ok, fixed.

content/articles/2017-11-01-lubridate-1.7.0.Rmd
```diff
@@ -0,0 +1,139 @@
...
+  author: Heather Zabriskie
+categories: [package]
+---
+
+```{r setup, include = FALSE}
+library("lubridate")
+knitr::opts_chunk$set(collapse = TRUE, comment = "##=>")
+```
+
+We are pleased to announce that `lubridate` 1.7.0 is now on CRAN! For a complete
```
hadley: * Just realised this needs a brief description of lubridate for those who aren't familiar with it. e.g. lubridate makes working with dates and times less painful by X, Y, and Z.

* And house style is format package names as regular names (i.e. no backticks)

Once that's done, I'll merge.

vspinu: > Just realised this needs a brief description of lubridate

I have added a link to CRAN's lubridate page. That should be enough. Lubridate is among the oldest packages in tidyverse, people surely heard about it. It's an announcement  of a new version targeted for people who use it, not an introduction of some sort. None of the earlier articles have added such descriptions either.

>  (i.e. no backticks)

Fixed.

hadley: There are lots of people new to the tidyverse and I think its a really good idea to include a very brief description. If previous blog posts didn't do it, it's only because I missed it.

vspinu: >  really good idea to include a very brief description.

Ok, I disagree, but adding it because it's your house and your rules ;-) I believe such post should be brief and to the point, excessive verbosity is distractive. One-click-away-description is a perfect trade-off IMO. 

BTW, I am considering rewriting lubridate from scratch and extracting the parser with some major simplifications. Hopefully within 6 months, but no promises.

content/articles/2017-11-01-lubridate-1.7.0.Rmd
```diff
@@ -0,0 +1,151 @@
...
+1.7.0](https://cran.r-project.org/web/packages/lubridate/index.html) is now on
+CRAN! For a complete set of changes please see the release
+[NEWS](https://github.com/tidyverse/lubridate/releases/tag/v1.7.0).
+
+Functions to work with date-times and time-spans: fast and user friendly parsing
+of date-time data, extraction and updating of components of a date-time (years,
+months, days, hours, minutes, and seconds), algebraic manipulation on date-time
+and time-span objects.
+
+Lubridate is a package that makes working with date-time and time-span objects
```
hadley: I think you accidentally duplicated this paragraph. Otherwise, it's good to go!

vspinu: Arh, sorry about that. Fixed. 

BTW, should titles be capitalized even if it's a package name? If the package name starts a sentence it would be capitalized. I think same should hold for titles, even when it's as short as "Lubridate 1.7.0". It  looks nicer this way on top of the page.

content/articles/2017-11-01-lubridate-1.7.0.Rmd
vspinu: >  some risk averse organisations use RHEL

And here we [go](https://github.com/tidyverse/lubridate/issues/601).

content/articles/2017-11-01-lubridate-1.7.0.Rmd
hadley: I don't have strong feelings either way, but to date we've treated package names like actual names and preserved capitalisation.

hadley:
Thanks! 

Do you want to tweet it or should I? (I think maybe you don't have a twitter account?)

vspinu:
Orh. I was just in the middle of writing the message for you to hold on on it. 

Brian wants to pull it of CRAN because of the missing CCTZ citation in the description :(  I am preparing a quick patch now.

vspinu:
There have been extra hiccups with rev-deps on solaris. The patch has been finally accepted and if Brian Ripley doesn't veto it again I will tweet about it tomorrow.

vspinu:
[Here](https://twitter.com/spinuvit/status/926812012499390465). Would appreciate if you could retweat. I have a micro-reach :) Thanks!

hadley:
Done!

PR #50, opened 2017-11-01
Upcase DataCamp

justmytwospence:
Makes it more obvious that DataCamp is a proper noun to people who aren't familiar.


jennybc:
Thanks!

PR #53, opened 2017-11-11
Added new course under 2017

leafyoung:
I just started a new course about Tidyverse packages this year. So I would like to add it here.


hadley:
Thanks!

PR #35, opened 2017-08-28
🍾💻🚗 Add Lucy's internship post (closes #23 🏀)

LucyMcGowan:
**TODO**
- [x] needs a better title 🙈 
- [x] what category should it be?
- [x] Should I crop my face out of the commits? I feel like there is way too much of my face, but without it they don't quite look like commits anymore 🤔. 

@jennybc & @hadley do you mind taking a :eyes: - I don't think I can request a review (maybe because I don't have write access?) 

😱 I JUST NOTICED that the issue this closes is #23, which is Michael Jordan's number, which ties in to dribbling 🏀 + space jam. Ergo, EVERYTHING IS META AND [AWESOME](https://m.youtube.com/watch?v=StTqXEQ2l-Y) 🎉

PR #55, opened 2017-11-15
🗂 describe convention used to add inline images

batpigandme:
_n.b._ I followed the method used in Lucy's internship post (PR https://github.com/tidyverse/tidyverse.org/pull/35), which may have been more complicated than necessary, given I didn't need to center the images...
¯\\(°_o)/¯


README.md
```diff
@@ -85,3 +85,11 @@ img %>%
   image_scale("1000x200") %>% 
   image_write("content/articles/googledrive-initial-release-wd.jpg")
 ```
+
+### Inline images
+
+Inline images can be added to articles by placing them in the `/static/images/` 
+directory. By current convention, these go in a subdirectory for each post.
+Images are added to the markdown using html, e.g. 
+`<img src="/static/images/subdir/image-name.jpg" />`.
```
hadley: I think you can still use markdown, the key is the full url

hadley:
I think maybe you pushed the 2 s hang to the wrong branch?

batpigandme:
The 2 second hang is on the live site
https://www.tidyverse.org/articles/2017/11/tidyverse_1.2.0/

README.md
batpigandme: Just gave this a whirl, and I think it's not working in markdown because of the config:
https://github.com/tidyverse/tidyverse.org/blob/cd7dc18c933738248775229be46127ec69eae591/config.toml#L2

hadley: It worked for me in the testthat 2.0.0 post. Can you please double check?

hadley:
Can you please ask @jimhester for a bit of help to get this branch into a better state? (i.e. just showing the changes to the README)

README.md
batpigandme: You're right (shocking, I know)— the problem was that I was using `"/images/blah/blah.gif"` with the quotation marks! 🙈

batpigandme:
Because I burned everything to the ground (which I promise not to do anymore), @jimhester says this can't be changed…

You can totally send me to the dreaded Lubyanka. I'm ready for it. It's a fitting punishment.

hadley:
Hmmm, in that case might be easiest to close this PR and just copy and paste the changes into a new PR?

jimhester:
We opened https://github.com/tidyverse/tidyverse.org/pull/70

hadley:
Oh, and that's what @jimhester has done (except that he would never stoop to copy and paste)

PR #56, opened 2017-11-15
Add withr 2.1.0 post

jimhester:
I am not entirely sure about some of the examples. I used plot examples for a little more visual pop, but maybe they are distracting and I should just do a text only example?

I also write to a pdf in one example, not sure how to handle that, I don't think putting it in the articles' directory is the right place.


jimhester:
Also I could use ggplot2 for the plot examples, I didn't mainly because I didn't want to add ggplot2 as a `Suggested` dependency in the package, since this post is also going to be a package vignette.

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+
+```{r setup, include = FALSE}
+knitr::opts_chunk$set(
+  comment = "#>",
+  collapse = TRUE
+)
+library(withr)
+```
+
+[withr 2.1.0](http://withr.r-lib.org) is now available on CRAN!
```
hadley: Link to release notes?

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+[withr](http://withr.r-lib.org) makes working with global state in R safer and
+less error prone. It also has no non-base dependencies so is easy to include in
+packages.
+
+Install the latest version with:
+```{r, eval = FALSE}
+install.packages("withr")
+```
+
+# Why withr?
```
hadley: I really wish this could be "Whither withr"

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+my_plot <- function(new) {
+  old <- par(col = "red", pch = 19)
+  plot(mtcars$hpp, mtcars$wt)
+  par(old)
+}
+my_plot()
+par("col")
+```
+
+Using the base function `on.exit()` is the robust solution to this problem,
```
hadley: the -> a

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+which will run the resetting code when the block is exited, regardless if it
```
hadley: which -> as `on.exit()`; regardless if -> regardless of

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+  plot(mtcars$hpp, mtcars$wt)
+}
+my_plot()
+par("col")
+```
+
+However this solution is somewhat cumbersome to work with. You
+need to remember to use an `on.exit()` call after each stateful call. In
+addition by default each `on.exit()` action will overwrite any previous
+`on.exit()` action in the same block (using `add = TRUE` prevents this). It is also
```
hadley: In parenthetical, mention even with `on.exit()` the calls get cleaned up in the wrong order?

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+  with_par(list(col = "red", pch = 19),
+    plot(mtcars$hp, mtcars$wt)
+  )
+  par("col")
+}
+my_plot()
+par("col")
+```
+
+In addition to the `with_*` functions there are `local_*` variants which simply
```
hadley: which simply -> whose effects

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+last until the end of the block they are included in. These work similar to
```
hadley: Is block technically the correct term? Maybe should be function?

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+`on.exit()`, but you can set the options in one call rather than two.
+
+```{r}
+par("col" = "black")
+my_plot <- function(new) {
+  local_par(list(col = "red", pch = 19))
```
hadley: I kind of wish these took `...` instead of requiring a list

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+Here we create a writable file connection, write some lines to it with
+`with_connection()`, then open a read-only connection and read the file using
+`local_connection()`.
+
+```{r}
+with_connection(list(con = file("temp", "w")), {
+  writeLines(c("foo", "bar"), con)
+})
+read_temp <- function() {
+  readLines(local_connection(file("temp", "r")))
```
hadley: I think `con <- local...` would make this slightly easier to read

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+}
+read_temp()
+```
+```{r, echo = FALSE}
+unlink("temp")
+```
+
+`with_db_connection()` provides [DBI](http://rstats-db.github.io/DBI)
+connections to Relational Database Management Systems which automatically call
```
hadley: Just databases is adequate I think

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+`DBI::dbDisconnect()`. Here we create a new [SQLite](https://www.sqlite.org/)
+database, connect to it with `with_db_connection()`, and write a new table to
+it. We then create another connection with `local_db_connection()` and read
+from the table.
+
+```{r}
+db <- tempfile()
+with_db_connection(
+  list(con = DBI::dbConnect(RSQLite::SQLite(), db)), {
```
hadley: Maybe use `:memory:` for a temporary in memory database that you don't need to clean up?

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+Three new functions allow you to temporary attach a package, namespace or
+environment in a block of code.
+
+```{r}
+with_package("lattice", {
+  xyplot(y ~ x, data.frame(x = -2:2, y = dnorm(-2:2)))
+})
+```
+
+`with_namespace()` attaches the package namespace, so all objects (including
```
hadley: Should we be promoting this?

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,234 @@
...
+
+There are many other bug fixes and other minor improvements in this release.
+You can see a complete list in the [release
+notes](https://github.com/tidyverse/withr/releases/tag/v2.1.0).
+
+A big thanks goes to all the community members who contributed code and opened
+issues since the last release!
+
+```{r, eval = FALSE, include = FALSE}
+x <- gh::gh("/repos/:owner/:repo/issues", owner = "r-lib", repo = "withr", since = "2017-06-16", state = "all")
```
hadley: Does this include PRs?

jimhester: Yes it should (https://developer.github.com/v3/issues/#list-issues)

>  As far as the API is concerned, every pull request is an issue, but not every issue is a pull request.

content/articles/2017-11-15-withr-2.1.0.Rmd
jimhester: Maybe not, I can remove this part.

content/articles/2017-11-15-withr-2.1.0.Rmd
jimhester: I think the issue there was I thought the memory database would be removed when the connection ended? Or does it persist until the R session ends?

content/articles/2017-11-15-withr-2.1.0.Rmd
jimhester: Unfortunately supporting this makes the implementations much more complicated, you need to then check which arguments are named, make sure there is only one etc. There also could be name clashes with regular arguments for things like `with_options()`.

While I agree the syntax is slightly more cumbersome than otherwise, I think the added consistency and simplicity makes it worthwhile.

content/articles/2017-11-15-withr-2.1.0.Rmd
jimhester: I technically can be any block environment

``` r
options("test" = 1)
local({
withr::local_options(c("test" = 2))
getOption("test")
})
#> [1] 2
getOption("test")
#> [1] 1
```

But this distinction is not really relevant to most people, I will change it to function.

content/articles/2017-11-15-withr-2.1.0.Rmd
jimhester: It can, even if it is not quite correct!

content/articles/2017-11-15-withr-2.1.0.Rmd
hadley: Awesome - this is way simpler than my code! @batpigandme

content/articles/2017-11-15-withr-2.1.0.Rmd
hadley: Doesn't block usually imply just `{}`?  i.e. your example wouldn't work without `local()`. Maybe environment is a better word?

content/articles/2017-11-15-withr-2.1.0.Rmd
hadley: Yeah, I understand, I just still wish they took `...` 😉

content/articles/2017-11-15-withr-2.1.0.Rmd
hadley: Oh I understand - you're using the same db across the two examples

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,226 @@
...
+  old <- par(col = "red", pch = 19)
+  plot(mtcars$hp, mtcars$wt)
+  par(old)
+}
+my_plot()
+par("col")
+```
+
+However one issue when there is an error before you are able to reset the
+options.
```
jennybc: "However this approach can fail if there's an error before you are able to reset the options."

content/articles/2017-11-15-withr-2.1.0.Rmd
```diff
@@ -0,0 +1,226 @@
...
+  plot(mtcars$hpp, mtcars$wt)
+  par(old)
+}
+my_plot()
+par("col")
+```
+
+Using a base function `on.exit()` is the robust solution to this problem.
+`on.exit()` will run the resetting code when the block is exited, regardless of it
+exits normally or with an error.
```
jennybc: "regardless of **whether** it exits ..."

jennybc:
I think this package is so handy, so I'm glad it's getting its own post!

Just in case it's interesting, other use cases I have special fondness for: fiddling with `.libPaths()` (esp. for `install_github()`), temporarily setting env vars, and just in testing more generally.

content/articles/2017-11-15-withr-2.1.0.Rmd
jimhester: Thanks! I think this frankenstein sentence was the result of a revision gone bad.

content/articles/2017-11-15-withr-2.1.0.Rmd
jimhester: Yeah, this is a limitation of `local()`, `on.exit()` (which is what I was thinking of) does work on all blocks

``` r
options(test = 1)
{
  print(getOption("test"))
  on.exit(options(test = 2))
}
#> [1] 1
getOption("test")
#> [1] 2
```

Anyway I think the distinction is not relevant for most uses, and removed the uses of block in favor of function.

jimhester:
Ok I think I have fixed everything, thank you both for the review!

jimhester:
I updated the date to be tomorrow, I _believe_ this means it will not be visible publicly until then if we merge it now, but am not positive.

jennybc:
Confirming that I don't see the post yet on <https://www.tidyverse.org/articles/>. So @jimhester seems to be correct re: consequences of dating the post. We better write that down.

hadley:
I looked through the blog code and I didn’t see anything.

PR #70, opened 2017-12-19
Add convention for including inline article images.

jimhester:
cc @batpigandme

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ img %>%
 
 Inline images can be added to articles by placing them in the `/static/images/` 
 directory. By current convention, these go in a subdirectory for each post.
-Images are added to the markdown using html, e.g. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,3 +85,11 @@ img %>%
   image_scale("1000x200") %>% 
   image_write("content/articles/googledrive-initial-release-wd.jpg")
 ```
+
+### Inline images
+
+Inline images can be added to articles by placing them in the `/static/images/` 
+directory. By current convention, these go in a subdirectory for each post.
+Images are added to the markdown using html, e.g.
```
batpigandme: Change the last line to:
Images are added to the markdown by giving the full file-path, e.g.

```
![](/images/subdir/image-name.jpg)
```

PR #79, opened 2018-01-04
Rebuild to eliminate Read More link in articles index

### LINKED ISSUE #76: Remove renegade read-more 🔗 on articles page
Possibly related to https://github.com/tidyverse/tidyverse.org/commit/e8fcdfbfea697ea011c8280b423da3d296e1b5f7 (though I'm not sure how), but there's a Read More link at the bottom of the [articles](https://www.tidyverse.org/articles/) page that goes to an un-styled version of Jenny's workflow post at <https://www.tidyverse.org/articles/1/01/>.

Visual depiction:
image

batpigandme:
* Fixes #76 
* All commits from 2739e0abed8947a38980634229101550bcefab59 onward (minus Jenny's workflow post commits)
* Re-introduces images and article for Jenny's workflow post, without the Read More link
* Builds correctly here: https://cranky-cori-3407d0.netlify.com/articles/


hadley:
The diff looks very suspicious - there shouldn't be so many changes. @jimhester can you plese help get a cleaner diff?

batpigandme:
@hadley they're all the commits made between the workflow post and present. The ones that are me alone are because they were my commits (and then bringing back the workflow post at the end). I think the link was a weird artifact (which I'd chalk up to the use of "naked" emoji— as in emoji w/out emo) from those commits. 

Unfortunately, I wasn't able to figure out a way to change things by modifying a source file, as the source file doesn't actually match the artifact files, which are regenerated even without the workflow post there using any build _after_ 2739e0abed8947a38980634229101550bcefab59.

jimhester:
What is the actual change we care about here, and what commit(s) is it coming from?

batpigandme:
@jimhester trying to get rid of the Read More link at the bottom of the articles index (see pretty picture in #76). This branch is off of e8fcdfbfea697ea011c8280b423da3d296e1b5f7, after which I wasn't able to build without getting the link.

This is me attempting to follow Jenny's instructions here: https://github.com/tidyverse/tidyverse.org/issues/76#issuecomment-355332393

jimhester:
What is the actual change (i.e. what lines in the diff remove this link) and what actual commits in the ones in the PR did it come from?

batpigandme:
@jimhester that's what I wasn't able to figure out. The rendered post, the one that shows up in `public > articles > 1 > 01` _presumably_ came from `2017-12-workflow-vs-script.md`, but there's no point at which it renders _only_ to the incorrect location (it always shows up in the right place as well). 

The first commit in which it shows up is 4d600b7e0971182f51600a44e49e3aa80818800b. If I delete the workflow post in a branch off of that commit it goes away, but that's not the case for any commits after which the workflow post was merged.

batpigandme:
Sorry, Jim. Just realized I didn't answer your second question which is:
cba5de612c9a1baf03a5e86ae0802970f92d2324
51fbf94637bfb0af7367de7d781318ff9f80c48a
0e7631f69d78b4de52f6227836e38a1fd7b275e2

jimhester:
So pretty sure the issue here is two fold.

1. The date on the post `2017-12-12` is ambiguous, as the 12 can either be the month or the day. If Hugo cannot parse the date it sets it to [0001-01-01](https://discourse.gohugo.io/t/why-is-publishdate-set-to-0001-01-01-00-00-00-0000-utc/1767) hence the strange entry in the index.
2. @jenny accidentally included a html file for an md only post, which generated two different posts. This HTML file is what you get linked to by the phantom post.

So the solution should be to remove the 2017-12-workflow-vs-script.html, and then regenerate the article index.

batpigandme:
Wow, @jimhester, you, sir, are a very good data detective… 🕵️
I did learn a lot about cherry picking, though. So, y'know, win-win. 😬

batpigandme:
Alas, this actually doesn't remove the link! I think it's because it's in public (which is in `.gitignore`, and, thus, doesn't get removed by removing the source file).

jimhester:
You need to force rebuild the article index, e.g. by running `hugo` on the command line.

batpigandme:
Oh, OK. Is there a way to do that through netlify? https://deploy-preview-79--tidyverse-org.netlify.com/

jimhester:
It looks to be fixed now.

batpigandme:
Awesome!

jennybc:
This was epic. Sorry for the various problematic features of my post.

The one thing we really can't blame on me: So ... we should make no posts on Jan 1, Feb 2, Mar 3, Apr 4, etc 🤔

jennybc:
@batpigandme Are you in a good position to emo-ify my post, just to bring everything up to a better standard? I think most of us consult existing content when writing new content, so it's probably good to model the behaviours what we want more of. If you can't, I can do it later today probably.

jimhester:
It seems like the date only became an issue because in the html file, the markdown date seemed to be parsed fine. ¯\_(ツ)_/¯

batpigandme:
@jennybc mos def can get the emojification sorted! And, good call re. the confusing dates!

jennybc:
Is this what we've learned:

  * If your post is `.Rmd`, check in `.Rmd` and `.html`.
  * If your post is `.md`, only check that in. No `.html`.

??

jennybc:
OK, then we need to update the README.

jimhester:
Did you mean to close this? were you planning to handle it differently?

batpigandme:
😱 @jimhester totally did not mean to do that!

Thanks for the heads up.

jennybc:
I think this is good to go, yes? I think I've been assuming that @hadley will press the button in this case 🙂

PR #81, opened 2018-01-05
Add lifecycle badge page

### LINKED ISSUE #80: Add lifecycle badge definitions
See https://github.com/r-lib/usethis/issues/193#issuecomment-355624042
```
 #' * Experimental: very early days, a lot of churn in search of a good API. 
 #'   Not on CRAN. Might not go anywhere. Use with care. 
 #' * Maturing: API roughed out, but finer details likely to change. Will strive 
 #'   to maintain backward compatibility, but need wider usage in order to get 
 #'   more feedback. 
 #' * Dormant: not currently working on it, but plan to come back to in the 
 #'   future. 
 #' * Stable: we're happy with the API, and the unlikely to be major changes. 
 #'   Backward incompatible changes will only be made if absolutely critical, and 
 #'   will be accompanied by a change in the major version. 
 #' * Questioning: no long convinced this is a good approach, but don't yet 
 #'   know what a better approach might be. 
 #' * Retired: known better replacement available elsewhere. Will remain 
 #'   available on CRAN. 
 #' * Archived: development complete. Archived on CRAN and on GitHub. 
```

batpigandme:
* Closes #80


content/lifecycle.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Lifecycle badges
+---
+
+![](/images/lifecycle/lifecycle.png)
+
+### Experimental ![](https://img.shields.io/badge/lifecycle-experimental-orange.svg) {#experimental}
+
+Very early days, a lot of churn in search of a good API.  Not on CRAN. Might not go anywhere. Use with care.
```
hadley: An experimental package is in the very early stages of development. The API will be changing frequently as we rapidly iterate and explore variations in search of the best fit. Experimental packages will make API breaking changes without deprecation, so you are generally best off waiting until the package is more mature before you use it. Experimental packages will not be released on CRAN

hadley: Do we need to explain what API is?

content/lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Maturing ![](https://img.shields.io/badge/lifecycle-maturing-blue.svg) {#maturing}
+
+API roughed out, but finer details likely to change. Will strive to maintain backward compatibility, but need wider usage in order to get more feedback.
```
hadley: The API of a maturing package has been roughed out, but finer details likely to change. Once released to CRAN, we will strive to maintain backward compatibility, but the package needs wider usage in order to get more feedback and find the optimal API.

content/lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Stable ![](https://img.shields.io/badge/lifecycle-stable-brightgreen.svg) {#stable}
+
+We're happy with the API, and the unlikely to be major changes. Backward incompatible changes will only be made if absolutely critical, and will be accompanied by a change in the major version.
```
hadley: In a stable package, we are largely happy with the API, and major changes are unlikely. This means that the API will generally evolve by adding new functions and new arguments; we will avoid removing arguments or changing the meaning of existing arguments.

If API breaking change are needed, they will occur gradually. To begin with, the function or argument will be deprecated; it will continue to work but will emit an message informing you of the change. Next, typically after at least a year, the message will be transformed to an error.

content/lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Retired ![](https://img.shields.io/badge/lifecycle-retired-orange.svg) {#retired}
+
+Known better replacement available elsewhere. Will remain  available on CRAN.
```
hadley: A retired package is no longer under active development, and a known better alternative is available. We will only make the necessary changes to ensure that retired packages remain on CRAN. No new features will be added, and only the most critical of bugs will be fixed.

content/lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Archived ![](https://img.shields.io/badge/lifecycle-archived-red.svg) {#archived}
+
+Development complete. Archived on CRAN and on GitHub.
```
hadley: The development of an archived package is complete, and it has been archived on CRAN and on GitHub.

content/lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Dormant ![](https://img.shields.io/badge/lifecycle-dormant-blue.svg) {#dormant}
+
+Not currently working on it, but plan to come back to in the future.
```
hadley: A dormant package is not completed, but is not currently under active development. We plan to return to it in the future.

content/lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Questioning ![](https://img.shields.io/badge/lifecycle-questioning-blue.svg) {#questioning}
+
+No longer convinced this is a good approach, but don't yet  know what a better approach might be.
```
hadley: We are no longer convinced that a questioning package is the optimal approach, but we don't yet know what a better approach is.

content/lifecycle.md
batpigandme: Great. Explain as in spell it out because it's first acronym use, or as in _explain_ explain?

hadley: EXPLAIN EXPLAIN

batpigandme: That seems like it would be kind of a lot of not-so-lifecycle-related content at the start…but, I'm also not a succinct explainer of things. Is there a good tidyverse definition of API somewhere?

hadley: Maybe an intro paragraph like:

This page describes the typical lifecycle on an R package. Knowing where a package is in its lifecycle is particularly important for understand how the API will change over time. The API (short for application programming interface) is the set of functions (and their arguments) that defines how you interact with the package. 

There are three ways an API can change:

1. It can __grow__ when a function or argument is added.
2. It can __shrink__ when a function or argument is removed.
3. It can __change__ when the meaning of an argument changes, or the type of data returned from a function changes.

Changing or shrinking the API will __break__ it: code that previously worked will no any longer. (Sometimes code will break even if the API doesn't; for example you might have accidentally depended on behaviour that the author thought was a bug)

batpigandme: Added.

content/lifecycle.md
```diff
@@ -3,10 +3,32 @@ title: Lifecycle badges
 ---
 
 ![](/images/lifecycle/lifecycle.png)
+### Package lifecycles & APIs {#api}
+
+This page describes the typical lifecycle on an R package. Knowing where a
+package is in its lifecycle is particularly important for understand how the API
```
jennybc: "for understandING"

batpigandme: Changed 👍

hadley:
Why don't we want to use svgs?

batpigandme:
I got confused. We do.

batpigandme:
As an export from graffle, or using the lifecycle badges themselves?

hadley:
Well now that you mention it, redrawing using the icons would be pretty awesome

batpigandme:
@hadley I can't get the lifecycle SVGs to open in graffle or illustrator for some reason. I was able to do the diagram, as you shared it with me (i.e. the one that's currently a png) as an svg. I put it in its own branch, just in case, but I can move the commit to this one, if you 👍 https://github.com/batpigandme/tidyverse.org/commit/2706d120c8b68f7e84aa2dd860aed9cc96a5698d

hadley:
Yeah, move it here

batpigandme:
Done

PR #84, opened 2018-01-08
Add lessons learned from #79 re. *.md vs *.Rmd

### LINKED ISSUE #82: Update README to reflect lessons learned in #79
https://github.com/tidyverse/tidyverse.org/pull/79#issuecomment-355638881

batpigandme:
* Closes #82
* No .html files with .md
* Use .Rmd for posts with emoji
* Use emo package to use emoji

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,10 @@ The source of the website is a collection of `.md` and `.Rmd` files stored in
   should be given name `year-month-slug.md`. Unfortunately this data isn't
   actually used when generating the output file: you'll need to set up 
   the yaml metadata. More on that below.  
-    + For `*.md` posts, no `*.html` file should be generated. If one is, it 
-    should not be committed/pushed, and can be deleted.  
+    + For `*.md` posts, no `*.html` file should be committed. If you generate one locally during development, delete it once it's no longer useful to you. Keep it out of this repo.
     
-    + For `*.Rmd` posts, an `*.html` file _should_ be generated when you run 
-    `blogdown::serve_site()` (see below).  
-    
-    + If you plan to include emoji in your post, you should use the `.Rmd` 
-    format, and generate emoji using the [emo](https://github.com/hadley/emo)
-    package.  
+    + For `*.Rmd` posts, an `*.html` file _should_ be committed. It will be generated when you run `blogdown::serve_site()` (see recommended workflow below).      
+    + If your post includes emoji, use the `.Rmd` format, and incorporate emoji using the [emo](https://github.com/hadley/emo) package.  
 
 * `data/events.yaml`: this yaml file contains information about upcoming 
   events. The site automatically filters out events that have happened,
@@ -43,7 +38,7 @@ The source of the website is a collection of `.md` and `.Rmd` files stored in
 
 ## Previewing changes
 
-### Locally
+### Use `blogdown::serve_site()`
 
 To build the site locally, you'll need to install blogdown, and then install 
 hugo, the music behind the magic of blogdown:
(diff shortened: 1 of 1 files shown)


jennybc:
I think this pre-existing sentence was always confusing and is more so now:

> The source of the website is a collection of .md files stored in content/, and turned in to html with blogdown.

We make several references to `.md` files but they aren't present in the repo for pages that originate with `.Rmd`, right? And for those files, we actually generate `.html` locally ourselves and commit it, right? I guess it's no surprise that I'm the one that goofed this up!

Here's my current understanding: We have two types of pages: those that are written as `.md` and those that are written as `.Rmd`.

  * If you are authoring in `.md`, do not commit any associated `.html` file.
  * If you are authoring in `.Rmd`, DO commit the `.html` file and the `.Rmd` file.

I think you @batpigandme should feel free to do some more invasive editing to clarify:

  * Which files get generated via blogdown for your local preview enjoyment vs. generated remotely for the actual site.
  * The workflow for pages based on `.md` vs `.Rmd`. As in, which files will get generated, where/by whom, and how to handle vis-a-vis git.

batpigandme:
@jennybc good points all around. Just a question: for me, when I did the workflow post as an `.md` file, it did _not_ generate an `.html` file locally. It _did_ generate one if I clicked **Preview** (I believe that was the button that appeared in place of **Knit**).

I don't think this changes the instructions (no matter what, we don't want the md ⇨ html file), I'm just curious as to whether or not this happens automatically.

jennybc:
It's very believable that I clicked that button, yes. So I think that's why people like me need absolute clarity around "generate" vs. "commit". So maybe you should mention the different ways of previewing (the Preview button vs. `blogdown::serve_site()`) and how they will differ re: locally generated files. Should I train myself not to click the button?

We will enjoy having clear instructions, plus other people will probably use this as a reference for their own projects.

batpigandme:
I think I read like half of a thread on RLadies wherein someone mentioned "don't click knit" re. blogdown, forever curing me of a desire to do so. However, I've been hitting preview and deleting the `.html` for editing the README. So, what you're saying makes perfect sense.

hadley:
We should make sure this confusion gets pushed upstream to blgodown. I didn’t get it either until I broke the rstudio blog (but in the opposition direction to this discussion - I gitignored *.html)

jennybc:
Do not let this bog down this PR: but I have wondered why we don't render `.Rmd` to `.md` and then let blogdown take it from there, as we do for pages with `.md` source?

batpigandme:
On my regular blogdown I don't do the `.md` source for the sole reason that it means I can't use footnotes!

batpigandme:
What I attempted to convey in words (hopefully more clearly than I've done in pictures):

**Not recommended, but it happens to the best of us:**
blogdown local preview

**Mandatory for rendering, best practice for previewing:**
blogdown rendered

README.md
```diff
@@ -9,24 +10,29 @@ This repo is the source of <https://tidyverse.org>, and this readme tells you ho
...
-The source of the website is a collection of `.md` files stored in [`content/`](content/), and turned in to html with [blogdown](https://bookdown.org/yihui/blogdown). 
+The source of the website is a collection of `.md` and `.Rmd` files stored in 
+[`content/`](content/), which are rendered for the site with 
+[blogdown](https://bookdown.org/yihui/blogdown). 
 
 * `content/*.md`: these files generate the top-level pages on the site:
   packages, learn, help, and contribute. 
   
 * `content/articles/`: these files are the tidyverse blog. New blog entries
   should be given name `year-month-slug.md`. Unfortunately this data isn't
```
jennybc: Should this be `year-month-slug.[R]?md` to stand in equally well for an `.md` and `.Rmd` post?

batpigandme: ✅ Yes.

PR #101, opened 2018-01-25
option() -> options()

krlmlr:
Closes tidyverse/tibble#376.


hadley:
That gets me every time too!

PR #118, opened 2018-02-26
Remake wide image for usethis 1.3.0 post

jennybc:
Trying to figure out why the wide image is not displaying on the usethis1.3.0 post.


hadley:
BTW the reason I don't automate this in my posts is that I like to choose different parts of the overall image.

jennybc:
Oh it’s not automated, believe me 😅. I just happen to record the magick code. Still not 100% sure why first image problematic but I suppose ... size?

PR #121, opened 2018-02-26
Blog post for pillar 1.2.1

krlmlr:
@hadley: This requires a hotfix to be pushed to CRAN, to fix lubridate. I think the coloring of negative decimal numbers is wrong too.

Preview link: https://5a943d98efbe5d6bd34765fe--tidyverse-org.netlify.com/articles/2018/03/pillar-1-2-1/.


hadley:
I think it might be easier for people to understand if you just use `tibble(x)` instead of `pillar(x)`

krlmlr:
Thanks. Would you agree to an explanatory note? It seems that *ansistrings* can currently render *pillar* output but not *tibble* output (r-lib/ansistrings#14), and I'd rather avoid taking screenshots again.

hadley:
Why is tibble not generated balanced escapes?

I'd really prefer to keep the blog post focussed on tibble, not on incidental details that most people won't care about.

krlmlr:
*tibble* uses nested *crayon* styles which are implemented with escape codes. I might be able to work around this in *tibble*.

krlmlr:
Changed examples to use *tibble*, added a clarification section to the bottom: https://deploy-preview-121--tidyverse-org.netlify.com/articles/2018/03/pillar-1-2-1/. Let's wait until Friday (the anticipated date of the blog post), binaries should be built by then.

lionel-:
FYI I needed this tweak to `colourise_chunk()`:

```r
colourise_chunk <- function(type) {
  function(x, options) {
    x <- pkgdown:::escape_html(x)
    x <- ansistrings::ansi_to_html(x, fullpage = FALSE)
    class <- tolower(options$engine)
    sprintf('<pre class="knitr %s">%s</pre>\n', class, x)
  }
}
```

It escapes HTML syntax which is necessary for printing stuff like `<quosure>`.

krlmlr:
Thanks. I'm already using `htmltools::htmlEscape(x)` in the call to `ansi_to_html()`, the code looks very similar to yours.

lionel-:
oops sorry I must have been confused while debugging and somehow got rid of `htmltools::htmlEscape()`

krlmlr:
All binaries have been built on CRAN, ready to publish?

hadley:
Published, thanks!

PR #126, opened 2018-03-05
Add analytics to tidyverse.org.

batpigandme:
* Add UA to `config.toml`.
* Add Google-Tag-Manager script to `head.html`.
* Closes #199.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -3,7 +3,7 @@ relativeurls = false
 languageCode = "en-us"
 title = "Tidyverse"
 description = "An opinionated collection of R packages designed for data science."
-googleAnalytics = ""
+googleAnalytics = "UA-115082821-1"
 disqusShortname = ""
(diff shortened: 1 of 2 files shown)


config.toml
```diff
@@ -3,7 +3,7 @@ relativeurls = false
 languageCode = "en-us"
 title = "Tidyverse"
 description = "An opinionated collection of R packages designed for data science."
-googleAnalytics = ""
+googleAnalytics = "UA-115082821"
```
hadley: It might be less confusing to remove this line altogether since our custom theme no longer uses it

batpigandme: Oh…ok. I thought GA and GTM were actually different. Will change.

hadley: Oh maybe they are?

hadley: Oh they are different. Do we need both?

PR #140, opened 2018-04-09
Add full content to RSS.

batpigandme:
Per @carneybill's request.
_n.b._ totally unformatted so, not sure how helpful this is.

Oddly, working in local preview, and _not_ on netlify deploy preview…
https://deploy-preview-140--tidyverse-org.netlify.com/articles/index.xml
vs
image


batpigandme:
@hadley is there a reason we don't use tidyverse.org as baseurl? Among other things, this would make the links in the RSS feed work.

https://github.com/tidyverse/tidyverse.org/blob/071acf8c1b403d227f01d6731ea4d7b882168240/config.toml#L1

batpigandme:
Works for real now!
image

hadley:
Probably because I forgot that base url will affect the embedded links and I assumed we wouldn't need it.

batpigandme:
I think that would only matter if we also changed `relativeurls` to `true`, which we shouldn't have to.

PR #147, opened 2018-04-24
Fix link to rlang package page.

### LINKED ISSUE #145: rlang.tidyverse.org returning 404
rlang.tidyverse.org is directing to a 404; it's possible this site has been relocated, in which case https://www.tidyverse.org/packages/ needs to be updated to reflect the new location

batpigandme:
* Fixes #145.
* _n.b._ Since rlang is now part of r-lib, should we remove it from the tidyverse package index entirely?


hadley:
I think we should just remove it altogether

batpigandme:
@hadley , good to merge?

PR #161, opened 2018-05-28
No tverse depends, imports, etc.

batpigandme:
Not sure if this should be an actual post, but, given new packages appear on the list, it might be worthwhile…


content/articles/2018-05-dont-depend-on-the-tidyverse.Rmd
```diff
@@ -0,0 +1,45 @@
+---
+title: Don't depend on the tidyverse
```
hadley: Maybe: "The tidyverse is for EDA, not packages" ?

PR #163, opened 2018-06-04
Conflicted 0.1.0

hadley:


content/articles/2018-06-conflicted.Rmd
```diff
@@ -0,0 +1,86 @@
...
+  author: Vincent van Zalinge
+tags: []
+---
+
+```{r}
+knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
+```
+
+
+We're stoked to announce the initial release of the [conflicted](https://github.com/r-lib/conflicted#conflicted) package. The goal of conflicted is to provide alternative way of resolving conflicts caused by ambiguous function names. R handles ambiguity by reporting conflicts when you load a package, then having the most recently loaded package win. This can make it hard to detect conflicts, because it's easy to miss the messages since you often load packages at the top of the script, and you don't see a problem until much later. conflicted takes a different approach to resolving ambiguity, instead making every conflict an error and forcing you to always choose which function to use.
```
batpigandme: * California Hadley leaves his mark with "stoked"... 🏄  (leave it, it's hilarious)
* _an_ alternative way
* _then having_ sounds a little weird to me (it matches with reporting, but... You could try "To handle ambiguity R reports..." or something of the like)
* Swap in explicitly for always?

content/articles/2018-06-conflicted.Rmd
```diff
@@ -0,0 +1,86 @@
...
+
+Installed conflicted by running:
```
batpigandme: Install (no ed)

content/articles/2018-06-conflicted.Rmd
```diff
@@ -0,0 +1,86 @@
...
+
+```{r}
+if (interactive()) {
+  require(conflicted)
+}
+```
+
+## How does it work?
+
+Loading conflicted creates a new "conflicted" environment that is attached just after the global environment. This environment contains an active bindings for any object that is exported by multiple packages; the active binding will throw an error message describing how to disambiguate the name.
```
batpigandme: No "an" before "active bindings" or no "s" on binding

content/articles/2018-06-conflicted.Rmd
```diff
@@ -0,0 +1,86 @@
...
+
+The conflicted environment also contains bindings for `library()` and `require()`, these  suppress the default conflict reporting (i.e. `warn.conflicts = FALSE`) and ensure that the conflicted environment is updated with any new conflicts. Because they occur in the search path before the base package, they will be called instead of `base::library()` and `base::require()`.
```
batpigandme: Either change these to which (these suppress the default...), or use a semicolon or start a new sentence at "these suppress"

PR #177, opened 2018-07-02
Render bench 1.0.1 post with TOC

batpigandme:


batpigandme:
Blurg! I have to figure out how to work around the diff. rendering of [`\@username`] with .Rmarkdown. From `.Rmd` this renders correctly, and _without_ the escape pandoc yells at you for not having a reference (thinking it's a citation). But, the markdown is rendered by Hugo with the `\` included…
image

batpigandme:
Proposal (decidedly a function of laziness), we could just nix the usage of `@`. Since it's on a webpage, and not GH, it doesn't actually connect to anything special... 
cc @hadley

hadley:
Another option would be to try `` `@blah` ``

batpigandme:
You like? 
image

hadley:
Hmmm, not really.

PR #188, opened 2018-07-09
Updated events 

carneybill:


batpigandme:
Bill, can you rebase your changes onto the current master? That way old events from 2017 won't be re-introduced? (i.e. it should take care of the merge conflicts).

batpigandme:
@hadley Do you want to take a peek? rstudioconf doesn't show up right now, but that will only be the case for three more days…

PR #211, opened 2018-08-28
Clarify wording on packages.md

angela-li:
Right now, it says "three" but there are only two packages mentioned after this sentence, so I changed the wording to be more clear (unless there's a secret third package beyond magrittr and glue that's not mentioned...)


batpigandme:
Thanks, Angela!

PR #213, opened 2018-09-13
Remove past event

batpigandme:
* Remove Max's workshop (already occurred), should make rstudio::conf show up again.


batpigandme:
Done

PR #222, opened 2018-10-25
WIP draft for ggplot2 3.1.0

batpigandme:
cc @clauswilke —

I can't tell if it has to do with my local setup or not, but I can't get the `geom_sf_label()` example working. I have to run out, but I can take a closer look later.


clauswilke:
What's the workflow here? How can I add on top of your commit? Do I fork your repo and make a PR to your repo?

clauswilke:
To fix your second example, read in the `nc` dataset using this line:
``` r
nc <- st_read(system.file("shape/nc.shp", package = "sf"), quiet = TRUE)
```

The other dataset has a column `geom` instead of `geometry` and needs an additional `aes(geometry = geom)` statement.

batpigandme:
Oh, shoot! Sorry, Claus. I forgot that I did this one in a fork! (And thanks for the fix — doing slides while writing posts is clearly not a good look for me).

Give me a minute, and I'll make a branch here so you can work off of it.

batpigandme:
OK @clauswilke, everything should be in the tidyverse branch now!

PR #225, opened 2018-10-25
Source the user rprofile after the project Rprofile

jimhester:
The site uses a project Rprofile, which means the user profile is not
evaluated, this change allows it to be evaluated if a user profile
exists.

This lead to some confusion on my part because my profile was not being executed until I figured out what was going on.

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -3,3 +3,9 @@ options(
   blogdown.rmd = TRUE,
   blogdown.subdir = "articles"
 )
+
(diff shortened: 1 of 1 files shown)


.Rprofile
```diff
@@ -3,3 +3,11 @@ options(
   blogdown.rmd = TRUE,
   blogdown.subdir = "articles"
 )
+
+if (interactive()) {
```
hadley: Shouldn't you always source the user profile?

jimhester: Yeah you are probably right

PR #231, opened 2018-11-14
generics 0.0.1

DavisVaughan:
* .md -> .Rmarkdown
* Edits from Mara
* Images


DavisVaughan:
Okay now it's ready!

batpigandme:
LGTM! Just for future reference, you can use `usethis::use_tidy_thanks()` (which automatically generates a list of the people who have contributed since whatever release you set it to for a given package) for the final section. But I like the way you have it here for this one!

PR #232, opened 2018-11-15
Add note about how to render .md from .Rmarkdown

DavisVaughan:
Currently there is no mention of `blogdown::serve_site()` as a way to preview the site, and no mention of how to render the `.md` from a `.Rmarkdown`. In retrospect this is simple, but as a first time poster I had no idea that I was supposed to do it.


batpigandme:
Yeah, it's actually in the README <https://github.com/tidyverse/tidyverse.org#previewing-changes>, but I don't think it hurts to have it in both places.

PR #239, opened 2018-11-26
Comments on parsnip post

alexpghayes:
cc @topepo. wasn't sure the best way to make comments but the post looks solid to me

### DIFF
--- a/content/articles/2018-11-parsnip-0-0-1.Rmarkdown
+++ b/content/articles/2018-11-parsnip-0-0-1.Rmarkdown
@@ -111,7 +111,11 @@ car_model <- linear_reg()
 car_model
 ```
 
-That's pretty underwhelming because we haven't given it any details yet. `parsnip` offers a variety of methods to fit this general model. We will use ordinary least squares, but could also use penalized least squares too (via the lasso, ridge regression, Bayesian estimation, dropout, etc). We differentiate these cases by the ***computational engines***, which is a combination of the estimation type, such as least squares, and the _implemention_. The latter could be an R package or some other computing platform like Spark or Tensorflow. 
+That's pretty underwhelming because we haven't given it any details yet. `parsnip` offers a variety of methods to fit this general model. We will use ordinary least squares, but could also use penalized least squares too (via the lasso, ridge regression, Bayesian estimation, dropout, etc).
+
+AH: Unclear to me what it means to fit a penalized linear model via dropout. maybe clarify this is casting the problem as a neural net.
+
+We differentiate these cases by the ***computational engines***, which is a combination of the estimation type, such as least squares, and the _implemention_. The latter could be an R package or some other computing platform like Spark or Tensorflow. 
 
 To start simple, let's use `lm`:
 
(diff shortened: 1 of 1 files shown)


content/articles/2018-11-parsnip-0-0-1.Rmarkdown
```diff
@@ -111,7 +111,11 @@ car_model <- linear_reg()
 car_model
 ```
 
-That's pretty underwhelming because we haven't given it any details yet. `parsnip` offers a variety of methods to fit this general model. We will use ordinary least squares, but could also use penalized least squares too (via the lasso, ridge regression, Bayesian estimation, dropout, etc). We differentiate these cases by the ***computational engines***, which is a combination of the estimation type, such as least squares, and the _implemention_. The latter could be an R package or some other computing platform like Spark or Tensorflow. 
+That's pretty underwhelming because we haven't given it any details yet. `parsnip` offers a variety of methods to fit this general model. We will use ordinary least squares, but could also use penalized least squares too (via the lasso, ridge regression, Bayesian estimation, dropout, etc).
+
+AH: Unclear to me what it means to fit a penalized linear model via dropout. maybe clarify this is casting the problem as a neural net.
```
topepo: Dropout is a form of regularization favored by the deep learning people. You can use it to fit a logistic or linear regression model.

content/articles/2018-11-parsnip-0-0-1.Rmarkdown
```diff
@@ -124,6 +128,8 @@ lm_car_model
 
 There are no additional arguments that we should specify here, so let's jump to fitting the actual model. Our two choices at this point are whether to use `fit()` or `fit_xy()`. `fit()` takes a formula, while `fit_xy()` takes objects for the predictors and outcome(s). Recall that `glm` and `lm` only allow for formulas, while `glmnet` only takes a matrix of predictors and an outcome. `parsnip` allows for either so that you can avoid having to think about what the underlying model function requires. To demonstrate, let's make a simple model:
 
+AH: perhaps comment on whether `fit_xy` needs matrices, or if it can also deal with data.frame/tibble
```
topepo: I want to stay away from adding a lot of detail in this post. To keep it concise, I've excluded a fair amount of low-level information.

content/articles/2018-11-parsnip-0-0-1.Rmarkdown
```diff
@@ -212,6 +218,8 @@ glmnet(x = x_mat, y = car_train$mpg, alpha = 0, lambda = 0.1)
 
 `alpha = 0` tells `glmnet` to only use an L2 penalty (as opposed to L1  and L2). 
 
+AH: the "possible syntax here make it unclear to me whether or not this actually works"
```
topepo: Do you mean that about the `glmnet` code above or the `keras` code below?

content/articles/2018-11-parsnip-0-0-1.Rmarkdown
alexpghayes: Right. This may be ignorance on my part, but for linear and logistic regression you have a convex objective with a single global optimum. Dropout doesn't change the objective, so I think that means that you're just doing coordinate descent on a random set of coordinates, and you should still converge to same the optimum unless you're doing some else like early stopping, etc. 

I've never seen someone do this and just thought it was a bit odd to suggest as a use case.

content/articles/2018-11-parsnip-0-0-1.Rmarkdown
alexpghayes: the `keras` code below

topepo: I'll make a note about that

content/articles/2018-11-parsnip-0-0-1.Rmarkdown
topepo: Hmm. I'll leave it in even though it isn't a common methdology

PR #266, opened 2019-01-06
Blog post for tibble 2.0.1

krlmlr:
I'm planning to add links to documentation for the funcitons mentioned in the text, unless there's some _pkgdown_ magic which I could use.


content/articles/2019-01-tibble-2-0-0.Rmarkdown
```diff
@@ -0,0 +1,187 @@
...
+fansi::set_knit_hooks(knitr::knit_hooks, which = c("output", "error"))
+```
+
+I'm pleased to announce that version 2.0.0 of the *tibble* package is on CRAN now. Tibbles are a modern reimagining of the data frame, keeping what time has shown to be effective, and throwing out what is not, with nicer default output too! Grab the latest version with:
+
+```r
+install.packages("tibble")
+```
+
+This release required a bit of preparation, including a [pre-release blog post](https://www.tidyverse.org/articles/2018/11/tibble-2.0.0-pre-announce/) that described the breaking changes.
```
hadley: I'd recommend briefly summarising the breaking changes here.

content/articles/2019-01-tibble-2-0-0.Rmarkdown
```diff
@@ -0,0 +1,187 @@
...
+- Robustness, lifecycle, name repair, row names, `glimpse()` for subclasses
+- _vctrs_, smaller footprint, decorations
+
+For a complete overview please see the [release notes](https://github.com/tidyverse/tibble/releases/tag/v2.0.0).
+
+Use the [issue tracker](https://github.com/tidyverse/tibble/issues) to submit bugs or suggest ideas, your contributions are always welcome.
+
+## Changes that affect users
+
+`tibble()` now supports columns that are matrices or data frames.
```
hadley: Mention that these have always been supported in data frames and are used in some modelling functions, but we're not yet sure how they will be used (but we want to be prepared).

This should probably come last in this section since it's the least impactful to most users.

hadley: You could also link to https://adv-r.hadley.nz/vectors-chap.html#matrix-and-data-frame-columns

content/articles/2019-01-tibble-2-0-0.Rmarkdown
```diff
@@ -0,0 +1,187 @@
...
+```{r enframe}
+enframe(letters[1:3])
+enframe(letters[1:3], name = NULL)
+```
+
+## Changes that affect package developers
+
+All functions have been assigned a lifecycle.
+The _tibble_ package has now reached the "stable" lifecycle, functions in a different lifecycle stage are marked as such in their documentation.
+One example is the `add_row()` function: it is unclear if it  should ensure that all columns have length one by wrapping in a list if necessary, adn a better implementation is perhaps possible once _tibble_ uses the _vctrs_ package, see below.
```
hadley:
```suggestion
One example is the `add_row()` function: it is unclear if it  should ensure that all columns have length one by wrapping in a list if necessary, and a better implementation is perhaps possible once _tibble_ uses the _vctrs_ package, see below.
```

content/articles/2019-01-tibble-2-0-0.Rmarkdown
```diff
@@ -0,0 +1,187 @@
...
+Therefore this function is marked "questioning".
+Learn more about lifecycle in the tidyverse at https://www.tidyverse.org/lifecycle/.
+
+The new `.rows` argument to `tibble()` and `as_tibble()` allows specifying the expected number of rows explicitly, even if it's evident from the data.
+This supports writing even more defensive code.
+The `nrow` argument to the low-level `new_tibble()` constructor is now mandatory, on the other hand most expensive checks have been moved to the new `validate_tibble()` function.
+This means that constructions of tibbles is now faster by default if you know that the inputs are correct, but you can always double-check if needed.
```
hadley: See motivation at https://adv-r.hadley.nz/s3.html#s3-classes

content/articles/2019-01-tibble-2-0-0.Rmarkdown
```diff
@@ -0,0 +1,187 @@
...
+## you can authorize duplicate names
+tibble(` ` = 1, ` ` = 2, .name_repair = "minimal")
+## or request that the names be made unique
+tibble(` ` = 1, ` ` = 2, .name_repair = "unique")
+## or or universal
+tibble(` ` = 1, ` ` = 2, .name_repair = "universal")
+```
+
+Row name handling is stricter.
+Row names are never (and never were) supported in `tibble()` and `new_tibble()`, and are now stripped by default in `as_tibble()`.
```
hadley:
```suggestion
Row names are not (and never were) supported in `tibble()` and `new_tibble()`, and are now stripped by default in `as_tibble()`.
```

content/articles/2019-01-tibble-2-0-0.Rmarkdown
```diff
@@ -0,0 +1,187 @@
...
+tools::package_dependencies("tibble", recursive = TRUE, which = "Imports")
+```
+
+Some of them, namely _fansi_ and _utf8_, are fairly heavy.
+[The plan](https://github.com/tidyverse/tibble/issues/475) is to make these packages, and _crayon_, a suggested package to _cli_, and provide fallback implementations there.
+When finished, taking a strong dependency on _tibble_ won't add too many new dependencies (again): _rlang_, _vctrs_ and _cli_ will be used by most of the tidyverse anyway, _pillar_ is the only truly new strong dependency.
+Packages that subclass `tbl_df` should import _tibble_ to make sure that the subsetting operator `[` always behaves the same.
+Constructing (subclasses of) tibbles should happen through `new_tibble()` only.
+
+Tibbles have a very opinionated way to print their data, not always in line with users' expectations, and sometimes clearly wrong (e.g. for numerical data where the absolute mean is much larger than the standard deviation).
```
hadley: I think it's probably worth mentioning there's a connection to vctrs here, as vctrs makes constructing new S3 vector classes much easier.

content/articles/2019-01-tibble-2-0-0.Rmarkdown
krlmlr: I wanted to say that no way of using `tibble()` or `new_tibble()` would yield a tibble with row names.?

hadley: Something is wrong with "are never" - maybe should be "were never"?

krlmlr:
@brodieG: Enabling colored HTML output using `fansi::set_knit_hooks(knitr::knit_hooks, which = c("output", "error"))` cripples some (but not all) tibbles in the output, see the preview at https://deploy-preview-266--tidyverse-org.netlify.com/articles/2019/01/tibble-2.0.0/. What am I doing wrong? Also, bold-faced ellipses are turned to a `…` character in some cases.

brodieG:
I can look at it more later, but at a glance looks like your chunk options are not right.  For example, you need `results="asis"` and `comment=""` when you set the hooks.  If you can, try to match what `vignette('sgr-in-rmd', package='fansi')` does exactly.  In particular [this](https://github.com/brodieG/fansi/blob/master/vignettes/sgr-in-rmd.Rmd#L62).  If that doesn't work let me know and I'll try to debug, but probably can't get to it till this evening US East Coast time.

krlmlr:
Thanks. This does change the output, but unfortunately doesn't help here:

screenshot from 2019-01-08 17-46-44

brodieG:
I tried just building the document with `rmarkdown::render` and it worked for me:

image

I'm guessing its a CSS conflict of some sort, unless maybe you don't have the latest version of `fansi`.  I'll have to download and build your whole site so I can see why it works as a standalone RMD file for me, but not for you.  The example above is a little modified to test the colors are actually being put in.

One other thing that would be helpful: can you report what `crayon::num_colors()` produces for you?

krlmlr:
_crayon_ reports 256 colors, I have _fansi_ 0.4.0 on my system.

The site is rendered using `blogdown::serve_site()`, you may need to touch the `.Rmarkdown` file to initiate the process.

I don't understand why the problem happens with some but not all tibbles in the output. I'll turn off colors as a stop gap, perhaps we can find a solution and update the blog post later.

krlmlr:
What's the timeline for merging and communicating this? 

There's no need to rush, binaries for some platforms haven't been built yet. I'm planning a 2.0.1 release fairly soon, but I'd like to wait before all binaries are built.

krlmlr:
OTOH, I think the binaries for R-release and R-devel will never be built due to the incompatibility with _rlang_ 0.3.1. Let's wait for _tibble_ 2.0.1 then.

brodieG:
Rename your file *.Rmd, not *.Rmarkdown.  I believe the latter triggers a blackfriday or some such markdown->html translation, which *#?*##! gratuitously substitutes HTML line breaks for new lines.  If you use Rmd I think you get the normal pandoc conversion directly to HTML.

I am not 100% certain this is what's happening because debugging the blogdown -> knitr -> rmarkdown -> pandoc chain (or any permutations of them) is about as pleasant and easy as pulling my nails out with wooden spoons and since the change was enough to make it work for me I do not want to spend one additional second confirming the exact nature of the failure.  However, this seems very likely because using Rmarkdown files leave .markdown artifacts, whereas using Rmd files leave html artifacts.

Please let me know if the change does not fix it for you.  I really really hope it does since I was both able to reproduce the problem and fix it by making the change.

hadley:
@krlmlr there's no rush, but at the same time, I don't think including coloured output is that important if it requires a very different setup to the rest of the blog.

@brodieG ah bummer — we deliberately use `.Rmarkdown` because if we use `.Rmd` we don't get an internal table of contents. Thanks for looking into it, and I 100% empathise with the pain of testing when you have this complicated combination of tools.

krlmlr:
I tweaked a bit for 2.0.1, should be ready to go.

Thanks again Brodie for looking into it, sorry for the pain. In the long run I'd love to see color in console output. Would it work if `fansi::set_knit_hooks()` generated HTML without newlines? A manual TOC or setting blackfriday options also seem worth investigating.

krlmlr:
Coloring works now, examples look good to me.

PR #269, opened 2019-01-25
Add blog post for purrr 0.3.0

lionel-:


content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+  fig.asp = 0.618, # 1 / phi
+  out.width = "700px"
+)
+
+library("purrr")
+```
+
+We're excited to announce the release of purrr 0.3.0! This release polishes the consistency of the interface, introduces new `pluck()`-based functions, many improvements to mappers, and a set of tools to limit the rate at which a function is called. Find a detailed account of the changes in the [NEWS](https://github.com/tidyverse/purrr/blob/master/NEWS.md#purrr-030) file.
+
+
```
jennybc: I think we should always reveal all the necessary code, having watched some people live code based on quick Googling. So that means always making this explicit:

  * Install the latest version via `install.packages("purrr")`.
  * Attach (or load?) purrr via `library(purrr)`.

content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+## Pluck assignment
+
```
jennybc: Maybe include a quick reminder of what `pluck()` is for? Quote or paraphrase from the docs:

> `pluck()` implements a generalised form of `[[` that allow you to index deeply and flexibly into data structures.

content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+The new functions `pluck<-`, `assign_in()` and `modify_in()` are assignment variants of `pluck()`. Let's create a nested data structure:
+
+```{r}
+x <- list(foo = list(1, 2), bar = list(3, 4))
+str(x)
+```
+
+This sort of repeated structures is the kind of data where `pluck()` shines:
```
jennybc:
```suggestion
This sort of repeated structure is the kind of data where `pluck()` shines:
```

content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+```
+
+You can now use the same syntax to modify the data:
+
+```{r}
+pluck(x, "foo", 2) <- 100
+str(x)
+```
+
+`pluck<-` also has a functional form that won't modify objects in your environment, but instead returns the modified data:
```
jennybc:
```suggestion
`pluck<-` also has a functional form that does not modify objects in your environment, but instead returns a modified copy:
```

content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+```
+
+Finally, `modify_in()` is a variant of `modify()` that only changes the pluck location with the result of applying a function:
+
+```{r}
+out <- modify_in(x, list("foo", 2), as.character)
+str(out)
+```
+
+assign_in() and modify_in()
```
jennybc: Seems to be leftover from writing process?

content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+```{r}
+map_at(x, -2, `*`, 1000)
+```
+
+
+## Consistency of behaviour
+
+A lot of effort went towards more consistent behaviour in purrr. An important improvement is that `modify()` is now a wrapper around `[[<-` rather than `[<-`. This should make it directly compatible with a larger set of vector classes.
+
+Thanks to the work of Mikko Marttila (@mikmart), `pmap()` and `pwalk()` do a better job of preserving S3 classes. `pluck()` also work better with S3 objects. In the next version of purrr, we plan to fix once and for all this sort of issues by integrating the vctrs package to deal with vector operations in a truly generic way.
```
jennybc:
```suggestion
Thanks to the work of Mikko Marttila (@mikmart), `pmap()` and `pwalk()` do a better job of preserving S3 classes. `pluck()` also works better with S3 objects. In the next version of purrr, we plan to use the in-development vctrs package to provide more principled and predictable vector operations.
```

To say "fix once and for all" just sounds like famous last words 😁

content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+
+```{r, eval = FALSE}
+reduce(1:4, `-`, .dir = "backward")
+
+compose(foo, bar, .dir = "forward")
+
+detect(1:5, ~ . %% 2 == 0, .dir = "backward")
+```
+
+Note that the backward version of `reduce()` (called right-reduce in the literature) applies the reduced function in a slightly different way that is more consistent with how this operation
```
jennybc: something's wrong with "with how this operation"

content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+### Stricter predicate checking
+
+purrr now checks the results of your predicate functions, which must now consistently return `TRUE` or `FALSE`. We no longer offer support for `NA` or for boolish numeric values (R normally interprets 0 as `FALSE` and all other values as `TRUE`). The purpose of this change is to detect errors earlier with a more relevant error messages.
+
+```{r, error = TRUE}
+keep(c(1, NA, 3), ~ . %% 2 == 0)
+```
+
+
+### Stricter pluck()
```
hadley: Move up with `pluck()`?

content/articles/2019-01-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,273 @@
...
+
+Thanks to Daniel Barnett (@daniel-barnett on Github), `pluck()` now has a stricter cousin `chuck()`. Whereas `pluck()` is very permissive regarding non-existing locations and returns `NULL` in these cases, `chuck()` fails consistently with informative messages:
```
hadley: (i.e. it "chucks" an error message)

content/articles/2019-01-purrr-0-3-0.Rmd
hadley: Also needs a brief overview of the rest of the documentation. It currently feels like a grab bag of different features; that's fine, but we should prepare the user for it.

lionel-: Instead of attach/load: "Start working with purrr"?

lionel-:
Reviews addressed, does this now look good to go?

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,308 @@
...
+```
+
+We're excited to announce the release of [purrr](https://purrr.tidyverse.org) 0.3.0! purrr enhances R’s functional programming toolkit by providing a complete and consistent set of tools for working with functions and vectors.
+
+```{r, eval = FALSE}
+# Install the latest version with:
+install.packages("purrr")
+
+# Start working with purrr:
+library("purrr")
```
hadley: I don't think our convention is use quotes here.

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,308 @@
...
+* Many missing functions were added to fill the gaps: `accumulate2()`, `imodify()`, `map_depth()`, ...
+
+* `partial()` has a much improved and more flexible interface.
+
+Find a detailed account of the changes in the [NEWS](https://github.com/tidyverse/purrr/blob/master/NEWS.md#purrr-030) file.
+
+
+## New pluck variants
+
+`pluck()` implements a generalised form of `[[` that allow you to index deeply and flexibly into data structures. For instance, `pluck(x, "foo", 2)` is equivalent to `x[["foo"]][[2]]`. purrr 0.3.0 introduces variants of `pluck()` to make it easier to work with deep data structures.
```
hadley: Probably worth mentioning the `.default` argument here too: For instance, `pluck(x, "foo", 2, .default = NA)` is equivalent to `x[["foo"]][[2]]`, returning an `NA` if that element doesn't exist.

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,308 @@
...
+
+
+### Pluck assignment
+
+This release introduces the new functions `pluck<-`, `assign_in()` and `modify_in()` as assignment variants of `pluck()`.
+
+To illustrate deep assignment, Let's create a nested data structure:
```
hadley: Can be in same para as previous.

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,308 @@
...
+
+```{r}
+out <- modify_in(x, list("foo", 2), as.character)
+str(out)
+```
+
+
+### Stricter pluck()
+
+Thanks to Daniel Barnett (@daniel-barnett on Github), `pluck()` now has a stricter cousin `chuck()`. Whereas `pluck()` is very permissive regarding non-existing locations and returns `NULL` in these cases, `chuck()` fails consistently with informative messages (i.e., it "chucks" an error message):
```
hadley: Whereas `pluck()` ..., and `[[` inconsistently returns `NULL`, `NA` or throws an error, `chuck()` ...

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,308 @@
...
+```
+
+`map_at()` now also supports negative selections:
+
+```{r}
+map_at(x, -2, `*`, 1000)
+```
+
+
+## Consistency of behaviour
```
hadley: I think you could make this heading just "Consistency", and then "Behaviour" could be a level 3 header.

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,308 @@
...
+```
+
+`slowly()` uses a constant rate by default while `insistently()` uses a backoff rate. The rate limiting can be configured with optional jitter via `rate_backoff()` and `rate_delay()`, which implement exponential backoff rate and constant rate respectively.
+
+```{r}
+walk(1:3, slowly(f, rate_backoff(max_times = Inf)))
+```
+
+
+## `map_if()`... or else?
```
hadley: These need a heading and a description in the overview (or just remove?)

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,308 @@
...
+# After
+exec(mean, 1:10, !!!list(na.rm = TRUE))
+```
+
+
+### Filling the missing parts
+
+* purrr 0.3.0 introduces `accumulate2()`, `modify2()` and `imodify()` variants.
+
+* By popular request, `at_depth()` is back as `map_depth()`. Unlike `modify_depth()` which preserves the class structure of the input tree, this variant only returns trees made of lists of lists, coercing vectors if needed.
```
hadley: Maybe move this up to next to `map_at()` and have a "map tweaks" level-2 header?

lionel-: I added a Map tweaks section but I think it's better to keep `map_depth()` here, so we can keep the subsection "Filling the missing parts".

lionel-:
@hadley now reorganised with new sections "Map and reduce improvements" and subsections "Behaviour" and "Interface" within "Consistency".

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,323 @@
...
+
+* The `rate_delay()` and `rate_backoff()` helpers control the invokation rate of `slowly()` and `insistently()`.
+
+The reduce and map functions gain a few improvements:
+
+* `map_if()` accepts an optional function with the `.else` parameter. This function is applied on elements for which the predicate is `FALSE`.
+
+* `map_at()` now accepts `vars()` selections. This lets you use selection helpers like `dplyr::starts_with()` to determine the elements of a list which should be mapped.
+
+* `reduce()` now support early termination of a computation. Just return a value wrapped in a `done()` to signal to `reduce()` that you're done.
```
jennybc:
```suggestion
* `reduce()` now supports early termination of a computation. Just return a value wrapped in a `done()` to signal to `reduce()` that you're done.
```

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,323 @@
...
+
+
+## New pluck variants
+
+`pluck()` implements a generalised form of `[[` that allow you to index deeply and flexibly into data structures. For instance, `pluck(x, "foo", 2)` is equivalent to `x[["foo"]][[2]]`. You can also supply a default value in case the element does not exist. For instance, `pluck(x, "foo", 2, .default = NA)` is equivalent to `x[["foo"]][[2]]`, returning an `NA` if that element doesn't exist. purrr 0.3.0 introduces variants of `pluck()` to make it easier to work with deep data structures.
+
+
+### Pluck assignment
+
+This release introduces the new functions `pluck<-`, `assign_in()` and `modify_in()` as assignment variants of `pluck()`. To illustrate deep assignment, Let's create a nested data structure:
```
jennybc:
```suggestion
This release introduces the new functions `pluck<-`, `assign_in()` and `modify_in()` as assignment variants of `pluck()`. To illustrate deep assignment, let's create a nested data structure:
```

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,323 @@
...
+`map_at()` now also supports negative selections:
+
+```{r}
+map_at(x, -2, `*`, 1000)
+```
+
+
+### Early termination of reduction
+
+‘reduce()’ is an operation that combines the elements of a vector into a single value by calling a binary function repeatedly with the result so far and the next input of a vector. `reduce()` and its variant `accumulate()` now support early termination of the reduction. To halt the computation, just return the last value wrapped in a `done()` box:
```
jennybc:
```suggestion
`reduce()` is an operation that combines the elements of a vector into a single value by calling a binary function repeatedly with the result so far and the next input of a vector. `reduce()` and its variant `accumulate()` now support early termination of the reduction. To halt the computation, just return the last value wrapped in a `done()` box:
```

Or maybe you really are trying to talk about "reduce" as a concept and not the function?

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,323 @@
...
+#### `modify()` now wraps `[[<-`
+
+An important improvement is that `modify()` is now a wrapper around `[[<-` rather than `[<-`. This should make it directly compatible with a larger set of vector classes.
+
+Thanks to the work of Mikko Marttila (@mikmart), `pmap()` and `pwalk()` do a better job of preserving S3 classes. `pluck()` also works better with S3 objects. In the next version of purrr, we plan to use the in-development vctrs package to provide more principled and predictable vector operations.
+
+
+#### Stricter predicate checking
+
+purrr now checks the results of your predicate functions, which must now consistently return `TRUE` or `FALSE`. We no longer offer support for `NA` or for boolish numeric values (R normally interprets 0 as `FALSE` and all other values as `TRUE`). The purpose of this change is to detect errors earlier with a more relevant error messages.
```
jennybc:
```suggestion
purrr now checks the results of your predicate functions, which must now consistently return `TRUE` or `FALSE`. We no longer offer support for `NA` or for boolish numeric values (R normally interprets 0 as `FALSE` and all other values as `TRUE`). The purpose of this change is to detect errors earlier with a more relevant error message.
```

content/articles/2019-02-purrr-0-3-0.Rmd
```diff
@@ -0,0 +1,323 @@
...
+
+```{r}
+my_list <- partial(list, lazy = rnorm(3), eager = !!rnorm(3))
+
+my_list()
+
+my_list()
+```
+
+You can also control the position of the future arguments by passing an empty `... = ` parameter. This syntax is powered by `rlang::call_modify()` to control where the position of dots in a function call:
```
jennybc: Something's off with this sentence / description of `rlang::call_modify()`.

lionel-: oh yes, also it's not easy to explain that syntax succintly. Does this do a better job?


> You can also control the position of the future arguments by passing an empty `... = ` parameter. This syntax is powered by `rlang::call_modify()` and allows you to add or move dots in a quoted function call. In the case of `partial()`, the dots represent the future arguments. We use this syntax in the following snippet to position the future arguments right between two partialised arguments:

lionel-:
Thanks for the reviews @jennybc and @hadley!

PR #273, opened 2019-02-12
Draft haven 2.1.0 post.

batpigandme:


batpigandme:
Note to self: Merge and update date once haven 2.1.0 is actually on CRAN

PR #286, opened 2019-03-21
Addition of  "Quick Starter Guides" section?

saghirb:
Proposal: Add section on "Quicker Starter Guides". Useful for the curious who want to get a flavour before investing more or to use as part of courses or workshops (e.g. for R Ladies/useR meetups).


hadley:
We're going to consider this page fully in the near future. Thanks for the suggestion!

PR #291, opened 2019-04-09
usethis v1.5.0

jennybc:
I still have a bit to do here, but it would be great to do that in parallel with feedback anyone else has.


jennybc:
I think this is fairly uncontroversial. Merging.

If you feel strongly about a modification, please let me know.

PR #296, opened 2019-04-10
Add event for the 1/2 day useR! tutorial on pkg dev

jennybc:
I think this makes sense to list here, but will defer to others if we think it somehow does not fit.


batpigandme:
Yeah, looks good. It won't actually show up until Hadley's workshop is done (we only have it laid out to show two - front-end design stuff Robby did with Hadley), but I think this is another good reason to use the ed team setup for a backup for events!

jennybc:
OK then. It hurts nothing to merge.

Sidebar: Feels like we should move on adopting some/all of the proposed innovations for this site.

batpigandme:
Yes, Allison and I are pencilled in to move on Monday b

PR #297, opened 2019-04-21
Testthat 2.1.0

hadley:


content/articles/2019-04-testthat-2-1-0.Rmarkdown
```diff
@@ -0,0 +1,112 @@
...
+
+```{r, eval = FALSE}
+install.packages("testthat")
+```
+
+testthat 2.1.0 mostly contains a large number of minor improvements and bug fixes, as described in the [release notes](https://github.com/r-lib/testthat/releases/tag/v2.1.0). This blog post discusses the bigger improvements:
+
+* `context()`s are now optional.
+
+* Two new expectations: `expect_invisible()` and `expect_mapequal()`.
```
batpigandme: Technically there's a list-agreement mismatch here (since the first is a sentence, and the second is a frag), but, since this matches headers, it's your call.

hadley: Yeah, I can't think of a better way to word it, and it's not too awful, so it stays

PR #306, opened 2019-05-06
Decrease line height in code blocks

### LINKED ISSUE #260: line height of code blocks
The line heights for code blocks feel really big to me, looks like it is set to 2x the font size (28px). This makes the code blocks really tall and IMHO you lose context with the rest of the post.

screen shot 2018-12-21 at 4 28 30 pm

Personally I find a `line-height: 18px` to look much cleaner, that is ~1.25x the font height.

screen shot 2018-12-21 at 4 27 59 pm

I don't suppose there is any chance of changing this...

jimhester:
Fixes #260

### DIFF
--- a/static/css/tidyverse-site.css
+++ b/static/css/tidyverse-site.css
@@ -912,7 +912,7 @@ pre, code {
 
 pre {
   padding: 10px;
-  line-height: 2em;
+  line-height: 1.5em;
   margin-bottom: 1em;
   word-wrap: normal;
   word-break: normal;
--- a/static/css/tidyverse-site.css
+++ b/static/css/tidyverse-site.css
@@ -912,7 +912,7 @@ pre, code {
 
 pre {
   padding: 10px;
-  line-height: 1.5em;
+  line-height: 1.5;
   margin-bottom: 1em;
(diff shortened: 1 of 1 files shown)


jimhester:
Sorry for the delay in doing the PR, I used 1.5, which tightens things up but is a still a bit bigger than what I initially suggested.

static/css/tidyverse-site.css
```diff
@@ -912,7 +912,7 @@ pre, code {
 
 pre {
   padding: 10px;
-  line-height: 2em;
+  line-height: 1.5em;
```
hadley: Should this just be `1.5`? I have a vague recollection that's safer.

jimhester: I think it has to have a unit unless it is `0` when (you are right) it is safest to omit it.

hadley: It's not a regular unit: https://css-tricks.com/almanac/properties/l/line-height/

jimhester: Oh, interesting! TIL, I have changed it to be unitless

Issue #307, opened 2019-05-06
table styling with raw `knitr::kable()` tables


jimhester:
The default table styling basically outputs tables without any padding or lines, it does not look great. Ideally this would give us nicer output of some kind.

```r
knitr::kable(mtcars)
```

jimhester:
I ended up with this css for a decently styled table

```css
td,th {
  padding: 0.4em;
}
thead {
border-top: 1px solid #aaa;
border-bottom: 1px solid #aaa;
}
table {
margin-left: auto;
margin-right: auto;
border-bottom: 1px solid #aaa;
}
```

PR #308, opened 2019-05-06
Add vroom blogpost

jimhester:


jimhester:
Looking at the output I think I should do something about how much output is generated for each code block. Either

1. Always assign the results to a variable rather than letting it auto-print.
2. Provide a specification or use `message = FALSE` to suppress the column type message.

Or maybe even both... thoughts?

jimhester:
@hadley, @jennybc or @batpigandme it would be great if one or more of you could review this!

batpigandme:
On it!

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+</html>
+
+I'm excited to announce that [vroom 1.0.0](http://vroom.r-lib.org) is now
+available on CRAN!
+
+vroom reads rectangular data, such as comma separated
+(csv), tab separated (tsv) or fixed width files (fwf) into R.
+
+It performs the
+same function as packages like [readr](http://readr.r-lib.org),
```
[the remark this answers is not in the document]
batpigandme: Also, maybe similar to instead of same as? (I'm just thinking about the scope of data.table).

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+[data.table](http://r-datatable.com) or the `read.csv()`
+function, but for many datasets vroom can read them much, much
+faster (hence the name).
+
+The main reason vroom can be faster is because character data is read from the
+file lazily; you only pay for the data you use. This lazy access is done
+automatically, so no changes to your R data manipulation code are needed.
```
batpigandme: hyphenate data-manipulation here (technically also R, R-data-manipulation, but I think that looks weird)

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+
+vroom also provides efficient multi-threaded writing that is multiple times
```
batpigandme: comma between efficient and multi-threaded

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+faster on most inputs than the readr writer.
```
batpigandme: Maybe `readr::write_*()` functions?

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+`delim` argument is used to specify the delimiter of the file explicitly.
+
+```{r}
+library(vroom)
+
+data <- vroom("flights.tsv")
+```
+
+The summary message after reading also differs from readr. We hope this output
+gives a more informative indication if the types of your columns are being guessed
```
batpigandme: "indication as to whether the types"?

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+install.packages("vroom")
+```
+
+The best way to get acquainted with the package is the [getting
+started](http://vroom.r-lib.org/articles/vroom.html) vignette.
+
+## Reading delimited files
+
+Compared to readr, the first difference you may note is you use only one
+function to read the files, `vroom()`. This is because `vroom()` guesses the
```
batpigandme: Link to function reference on first mention. [`vroom()`](http://vroom.r-lib.org/reference/vroom.html)

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+h2 code {
+    font-size: 1em;
+}
+</style>
+</html>
+
+I'm excited to announce that [vroom 1.0.0](http://vroom.r-lib.org) is now
+available on CRAN!
+
+vroom reads rectangular data, such as comma separated
```
hadley: I think you can combine these sentences into a paragraph.

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+```{r}
+library(vroom)
+
+data <- vroom("flights.tsv")
+```
+
+The summary message after reading also differs from readr. We hope this output
+gives a more informative indication if the types of your columns are being guessed
+properly. However you can still retrieve and print the full column specification
+with `spec()`.
```
batpigandme: Link to [`spec()`](http://vroom.r-lib.org/reference/spec.html) here

content/articles/2019-05-vroom-1-0-0.Rmarkdown
hadley: I think we need a couple of sentences about vroom vs readr somewhere.  i.e. we're not entirely sure yet, but we'll probably let them evolve separately for a little bit, but we plan to unite in the future. The major downside of vroom is that the laziness means that you can't get all problems up front, so unification will require some thought.

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+```
+
+The best way to get acquainted with the package is the [getting
+started](http://vroom.r-lib.org/articles/vroom.html) vignette.
+
+## Reading delimited files
+
+Compared to readr, the first difference you may note is you use only one
+function to read the files, `vroom()`. This is because `vroom()` guesses the
+delimiter of the file automatically (based on the first few lines). This works
```
hadley: Mention this is inspired by data.table

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+```
+
+## Reading multiple files
+
+One feature new to vroom is built-in support for reading sets of files with the
+same columns into one table. Just pass the filenames to be read directly to
+vroom.
+
+```{r}
+# Split the flights data by carrier
```
hadley: I'd hide this code, and instead say something like: "Imagine we have a directory containing ..."

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+vroom can also read files from the internet as well by passing the URL of the file to `vroom()`.
+
+```{r}
+file <- "https://raw.githubusercontent.com/r-lib/vroom/master/inst/extdata/mtcars.csv"
+data <- vroom(file)
+```
+
+It can even read gzipped files from the internet (although currently not the other compressed formats).
+
+## Reading and writing from pipe connections
```
hadley: Personally, I don't think this is important enough to include in the announcement, and including live code in data import makes me nervous.

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+- Select with the column names
+```{r}
+data <- vroom("flights.tsv", col_select = c(year, flight, tailnum))
+```
+- Drop columns by name
+```{r}
+data <- vroom("flights.tsv", col_select = c(-dep_time, -air_time:-time_hour))
+```
+
+- Use numeric indices
```
hadley: I'd also drop this as an example

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+s <- spec(data)
+
+data <- vroom("flights.tsv", col_types = s)
+```
+
+## Reading multiple files
+
+One feature new to vroom is built-in support for reading sets of files with the
+same columns into one table. Just pass the filenames to be read directly to
+vroom.
```
batpigandme: I think this last vroom should be `vroom()`?

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+```{r}
+vroom("flights.tsv", .name_repair = janitor::make_clean_names)
+
+vroom("flights.tsv", .name_repair = ~ janitor::make_clean_names(., case = "lower_camel"))
+```
+
+
+## Column types
+
+Like readr, vroom guesses the data types of columns as they are read, however sometimes it
```
hadley: Mention improved heuristic (i.e. looks at data throughout file, not just first `n` rows)

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+
+```{r}
+# Split the flights data by carrier
+purrr::iwalk(
+  split(flights, flights$carrier),
+  ~ vroom::vroom_write(.x, glue::glue("flights_{.y}.tsv"), delim = "\t")
+)
+```
+
+Then we can efficiently read all of the files into one tibble by passing the filenames
```
batpigandme: Comma after Then, here

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+data <- vroom("flights.tsv",
+  col_types = list(dest = col_factor(levels = c("EWR", "JFK", "LGA")))
+)
+```
+
+## Speed
+
+vroom is fast, but how fast?
+We benchmarked vroom using a real world dataset of taxi trip data, with
+14.7 million rows, 11 columns. It contains a mix of numeric and textual data and has a
```
hadley:
```suggestion
14.7 million rows, 11 columns. It contains a mix of numeric and text data and has a
```

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+```{r}
+bench::workout({
+  vroom_write(flights, "flights.tsv.gz")
+  vroom_write(flights, pipe("pigz > flights.tsv.gz"))
+})
+```
+
+## Column selection
+
+`vroom` introduces a new argument `col_select` which makes selecting columns to
```
batpigandme: appositive commas around `col_select`. So:
`vroom` introduces a new argument, `col_select`, which makes selecting columns…

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+
+  - Reading the data
+  - `print()`
+  - `head()`
+  - `tail()`
+  - Sampling 100 random rows
+  - Filtering for "UNK" payment, this is 6434 rows (0.0435% of total).
+  - Aggregation of mean fare amount per payment type.
+
+<style>
```
hadley: Add this to https://github.com/tidyverse/tidyverse.org/issues/307 ?

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+  vroom_write(flights, pipe("pigz > flights.tsv.gz"))
+})
+```
+
+## Column selection
+
+`vroom` introduces a new argument `col_select` which makes selecting columns to
+keep (or omit) more straightforward.
+
+The parameter uses the same interface as `dplyr::select()`, so you can do flexible selection operations.
```
batpigandme: I'd specifically state `col_select` again here, rather than "The parameter," just since it's a new paragraph.

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+
+```{r cleanup, include = FALSE}
+unlink(c(files, "flights.tsv", "flights.tsv.gz"))
+```
+
+<br/>
+
+Some things to note in the results. The initial reading is much faster in vroom
+than any other method, and most of the manipulations, such as `print()`,
+`head()`, `tail()` and `sample()` are equally fast. However because the
```
hadley: So fast you can't see them in the plot

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+character data is read lazily operations such as `filter()` and `aggregrate()`
+which need character values require additional time.
+However this cost will only occur once, after the values have been read they
```
hadley: So why are both "aggregate" and "filter" quite wide?

(You might also rename "aggregate" to "summarise" because I keep thinking you mean the base function)

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+* `col_time(format = "")` [t]: with the locale's `time_format`.
+* `col_datetime(format = "")` [T]: ISO8601 date times.
+* `col_factor(levels, ordered)` 'f', a fixed set of values.
+* `col_character()` 'c', everything else.
+* `col_skip()` '_, -', don't import this column.
+* `col_guess()` '?', parse using the "best" type based on the input.
+
+You can tell vroom what columns to use with the `col_types()` argument in a number of ways.
+
+If you only need to override a single column the most concise way is to use a named vector.
```
batpigandme: comma after single column

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+data <- vroom("flights.tsv", col_types = c(year = "i"))
+
+# also skip reading the 'time_hour' column
+data <- vroom("flights.tsv", col_types = c(year = "i", time_hour = "_"))
+
+# also read the carrier as a factor
+data <- vroom("flights.tsv", col_types = c(year = "i", time_hour = "_", carrier = "f"))
+```
+
+However you can also use the `col_*()` functions in a list.
```
batpigandme: comma after however

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+```{r}
+data <- vroom("flights.tsv",
+  col_types = list(dest = col_factor(levels = c("EWR", "JFK", "LGA")))
+)
+```
+
+## Speed
+
+vroom is fast, but how fast?
+We benchmarked vroom using a real world dataset of taxi trip data, with
```
batpigandme: real-world gets hyphenated here, since it's modifying dataset (technically taxi-trip, too, but dealer's choice there)

content/articles/2019-05-vroom-1-0-0.Rmarkdown
batpigandme: I'd put a comma after "a mix of numeric and textual data", and has... since you've got a list w/in list situation

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+```{r cleanup, include = FALSE}
+unlink(c(files, "flights.tsv", "flights.tsv.gz"))
+```
+
+<br/>
+
+Some things to note in the results. The initial reading is much faster in vroom
+than any other method, and most of the manipulations, such as `print()`,
+`head()`, `tail()` and `sample()` are equally fast. However because the
+character data is read lazily operations such as `filter()` and `aggregrate()`
```
batpigandme: offset "which need character values" here with commas

batpigandme: Also, comma after lazily

content/articles/2019-05-vroom-1-0-0.Rmarkdown
batpigandme: For this sentence I'd suggest:
However, this cost will only occur once. After the values have been read, they will be stored in memory, and subsequent accesses will be equivalent to other packages.

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,450 @@
...
+will be stored in memory and additional accesses will be equivalent to other
+packages.
+
+For more details on how the benchmarks were performed and additional benchmarks
+with other types of data see the [benchmark
+vignette](http://vroom.r-lib.org/articles/benchmarks.html).
+
+## Feedback welcome!
+
+vroom is a new package and like any newborn may fall down a few times before
```
batpigandme: offset "like any newborn" with commas

content/articles/2019-05-vroom-1-0-0.Rmarkdown
jimhester: Yeah you are right, I forgot to include something like this.

content/articles/2019-05-vroom-1-0-0.Rmarkdown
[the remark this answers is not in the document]
jimhester: yeah totally right, will change

content/articles/2019-05-vroom-1-0-0.Rmarkdown
```diff
@@ -0,0 +1,452 @@
...
+}
+</style>
+</html>
+
+I'm excited to announce that [vroom 1.0.0](http://vroom.r-lib.org) is now
+available on CRAN!
+
+vroom reads rectangular data, such as comma separated
+(csv), tab separated (tsv) or fixed width files (fwf) into R. It performs
+similar roles to functions like [readr::read_csv()](http://readr.r-lib.org),
```
batpigandme: Whoops — need backticks around `readr::read_csv()` and `data.table::fread()`

PR #323, opened 2019-07-09
Update link to Stanford Data Challenge Lab class

behrman:
Update link to curriculum for most current quarter.


behrman:
We now have a better URL to use. Will submit a new PR.

PR #343, opened 2019-10-01
new dials post for 0.0.3

topepo:


content/articles/2019-10-dials-0-0-3.Rmarkdown
```diff
@@ -0,0 +1,74 @@
...
+
+```{r p-set}
+boosting_set <- param_set(list(trees(), splits = tree_depth(), min_n()))
+boosting_set
+
+# modifying the parameter range:
+boosting_set %>% update(trees = trees(c(100, 1000)))
+```
+
+Note that the tree depth parameter has a user-defined identification variable. This cna come in handy when there are multiple tuning parameters of the same type. For example, suppose two variables (`x1` and `x2`) were modeled using splines. The flexibility of each grouped be represented in a parameter set:
```
batpigandme:
```suggestion
Note that the tree depth parameter has a user-defined identification variable. This can come in handy when there are multiple tuning parameters of the same type. For example, suppose two variables (`x1` and `x2`) were modeled using splines. The flexibility of each grouped be represented in a parameter set:
```

topepo: I was... um.. just checking to see if anyone would notice that. 😐

PR #344, opened 2019-10-01
typos

tmstauss:


hadley:
Thanks for the fix — unfortunately we forgot to merge it before reorganising the site, so now it's rather complicated.

PR #348, opened 2019-11-04
Remove past event

batpigandme:


hadley:
Once the new site is ready, we should dog food github actions by using it to re-build the site every day, so we don't have to worry about this any more

batpigandme:
@hadley, Yeah — Alison has events out of yaml format, so it can be date-dependent, but that'll work anyhow.

hadley:
I don't think it's the yaml that's the source of the problem; the problem is the site is only re-built when we (e.g.) publish a blog post, so the code that automatically filters the past events isn't run. IOTW you could actually fix this specific issue by modifying any file on the site because it would cause a rebuild.

batpigandme:
That makes sense. The thing that was happening previously is that past events weren't shown (as was the case just now), but you had to remove the old events in order to get multiple future events to show up.

Regardless, I understand what you're saying now re. rebuilding nightly.

PR #360, opened 2019-11-11
dtplyr 1.0.0

hadley:


content/blog/2019-11-dtplyr-1-0-0/index.Rmarkdown
```diff
@@ -0,0 +1,228 @@
...
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+I'm very excited to announce that [dtplyr](https://dtplyr.tidyverse.org) 1.0.0 is now on CRAN. dtplyr provides a [data.table](http://r-datatable.com/) backend for dplyr, allowing you to write dplyr code that is automatically translated to the equivalent, but usually much faster, data.table code.
```
batpigandme: > code that is automatically translated to the equivalent, but usually much faster, data.table code.

This resolves in a weird way where it's unclear if you're saying it's faster than data.table code or dplyr code. (I know it's the latter, but I guarantee it'll be parsed incorrectly). I think you can leave it as equivalent data.table code here, and leave the performance boost for later.

content/blog/2019-11-dtplyr-1-0-0/index.Rmarkdown
```diff
@@ -0,0 +1,228 @@
...
+
+```{r}
+mtcars2 %>% 
+  filter(wt < 5) %>% 
+  mutate(l100k = 235.21 / mpg) %>% # liters / 100 km
+  group_by(cyl) %>% 
+  summarise(l100k = mean(l100k))
+```
+
+Generally, however, you should reserve this for exploration and debugging, and instead use `as.data.table()`, `as.data.frame()`, or `as_tibble()` to indicate that you're done writing the transformation and want to access the results:
```
batpigandme:
```suggestion
Generally, however, you should reserve printing for exploration and debugging, and instead use `as.data.table()`, `as.data.frame()`, or `as_tibble()` to indicate that you're done writing the transformation and want to access the results:
```

batpigandme: Or some better word—just replace the "this."

content/blog/2019-11-dtplyr-1-0-0/index.Rmarkdown
```diff
@@ -0,0 +1,228 @@
...
+  filter(wt < 5) %>% 
+  mutate(l100k = 235.21 / mpg) %>% # liters / 100 km
+  group_by(cyl) %>% 
+  summarise(l100k = mean(l100k)) %>% 
+  as_tibble()
+```
+
+## Performance
+
+You might wonder how fast dtplyr is. That's a difficult question to answer in general (because there are so many different ways to use it), so here I'll focus on why dtplyr might be slower than data.table. There are three main reasons: translation cost, copies, and interface mismatch
```
batpigandme:
```suggestion
You might wonder how fast dtplyr is. That's a difficult question to answer in general (because there are so many different ways to use it), so here I'll focus on why dtplyr might be slower than data.table. There are three main reasons: translation cost, copies, and interface mismatch.
```

batpigandme: You could bring back the "generally faster" from the opening paragraph here re. comparison to equivalent dplyr code.

content/blog/2019-11-dtplyr-1-0-0/index.Rmarkdown
```diff
@@ -0,0 +1,228 @@
...
+
+### Translation cost
+
+Each dplyr verb must do some work to convert dplyr syntax to data.table syntax. We can use the [bench](http://bench.r-lib.org/) package to time the cost of four step pipeline that I used above:
```
batpigandme: Is it really the dplyr verb doing the work? I guess it wouldn't be "dtplyr verb" either. I can't think of anything better, but if you do...

batpigandme:
```suggestion
Each dplyr verb must do some work to convert dplyr syntax to data.table syntax. We can use the [bench](http://bench.r-lib.org/) package to time the cost of the four-step pipeline that I used above:
```

content/blog/2019-11-dtplyr-1-0-0/index.Rmarkdown
```diff
@@ -0,0 +1,228 @@
...
+    group_by(cyl) %>% 
+    summarise(l100k = mean(l100k))
+)
+```
+
+Because this pipeline does not use `as.data.table()` or `print()` it only generates the data.table code, it doesn't run it, so we're timing the translation cost. The translation cost scales with the complexity of the pipeline, not the size of the data, so these timings will apply regardless of the size of the underlying data.
+
+My intial experiments suggest that the translation cost is typically a few milliseconds. Since the computational cost increases with the size of the data, the translation cost becomes a smaller proportion of the total as the data size grows, suggesting the dtplyr does not impose a significant overhead on top of data.table.
+
+Take the following example, which uses large nycflights13 dataset. This isn't really big enough for data.table to really shine, but it's about as big as you can get in an R package. Here I'm going to compute the average arrival delay by destination. It takes raw dplyr about 60ms to do the work. Again, the dtplyr translation is fast, around 1ms, and then computation using data.table only takes about 30ms, almost twice as fast as dplyr.
```
batpigandme:
```suggestion
Take the following example, which uses the large nycflights13 dataset. This isn't really big enough for data.table to really shine, but it's about as big as you can get in an R package. Here I'm going to compute the average arrival delay by destination. It takes raw dplyr about 60ms to do the work. Again, the dtplyr translation is fast, around 1ms, and then computation using data.table only takes about 30ms, almost twice as fast as dplyr.
```

batpigandme: Or I guess it could be **a** large nycflights13 dataset—you just need an article.

content/blog/2019-11-dtplyr-1-0-0/index.Rmarkdown
```diff
@@ -0,0 +1,228 @@
...
+
+There is one place where dtplyr does have to add overhead: when translations would generate data.table code that modifies the input in place, like `mutate()`. dtplyr matches dplyr semantics (which never modifies in place), so most expressions involving `mutate()` must make a copy:
+  
+```{r}
+mtcars2 %>% 
+  mutate(x = x * 2) %>% 
+  show_query()
+```
+
+However, dplyr never generates more than one copy (no matter how many mutates you use), and it also recognises many situations where data.table creates an implicit copy:
```
batpigandme:
```suggestion
However, dtplyr never generates more than one copy (no matter how many mutates you use), and it also recognises many situations where data.table creates an implicit copy:
```

content/blog/2019-11-dtplyr-1-0-0/index.Rmarkdown
```diff
@@ -0,0 +1,228 @@
...
+  show_query()
+```
+
+You can learn more in [`vignette("translation")`](https://dtplyr.tidyverse.org/articles/translation.html). 
+
+There are a couple of limitations that I hope to address in the next version of dtplyr. Currently, you can't translate [the `_if` variants](https://github.com/tidyverse/dtplyr/issues/109), and there is weak support for the [`group_` functions](https://github.com/tidyverse/dtplyr/issues/108). If you discover other functions that don't work as you expect, [please file an issue!](https://github.com/tidyverse/dtplyr/issues/new/choose).
+
+## Acknowledgements
+
+Big thanks to the data.table community, particularly [Michael Chirico](https://github.com/MichaelChirico) for their help educating me on the best way to translate dplyr code into performant, idiomatic, data.table code.
```
batpigandme:
```suggestion
Big thanks to the data.table community, particularly [Michael Chirico](https://github.com/MichaelChirico), for their help educating me on the best way to translate dplyr code into performant, idiomatic, data.table code.
```

content/blog/2019-11-dtplyr-1-0-0/index.Rmarkdown
hadley: I went with "this preview"

PR #363, opened 2019-11-12
Show translation timings in benchmark

DavisVaughan:
@hadley I noticed that the blog post wasn't showing the translation benchmark because the expression is too long.

This also shows `allow.cartesian = TRUE` in the translations now, I think that was a new change right before release?

The benchmark timings cut in half, but maybe I have more cores on my machine than you. Relative differences are essentially the same.


hadley:
Thanks!

PR #367, opened 2019-11-14
Add `-s` flag to scheduled build.

batpigandme:
After looking at the past two nights builds (click **Actions** in the top menu to see), I'm not sure the build hook is actually _doing_ anything. 

I've been following the two posts (below), and I _think_ adding the `-d` flag should fix it, but, @jimhester, if you could take a look, I'd appreciate it!

https://ericjinks.com/blog/2019/netlify-scheduled-build/

https://www.voorhoede.nl/en/blog/scheduling-netlify-deploys-with-github-actions/

cc @apreshill — I tried to see what you're doing on the education site, but it doesn't look like there's an Actions tab.

### DIFF
--- a/.github/workflows/scheduled_build.yml
+++ b/.github/workflows/scheduled_build.yml
@@ -9,6 +9,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Trigger build hook on Netlify
-      run: curl -X POST "https://api.netlify.com/build_hooks/${TOKEN}"
+      run: curl -X POST -d "https://api.netlify.com/build_hooks/${TOKEN}"
       env:
         TOKEN: ${{ secrets.NETLIFY_CRON_BUILD }}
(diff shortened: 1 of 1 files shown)


jimhester:
Hmm, I don't think this will change anything, you aren't posting any data. Can you see on the Netlify builds pane if a build was triggered by the scheduled job?

batpigandme:
It was. So maybe there isn't a problem…
image

jimhester:
Yeah I think it is good then and you should be able to close this. Though maybe you can add `-s` to the curl command, since the progress bar output is not terribly useful.

jennybc:
Screen Shot 2019-11-14 at 8 17 32 AM

Maybe there simply weren't any changed files that needed to be deployed in these scheduled builds.

batpigandme:
@jimhester added `-s` and ditched `-d`

@jennybc That makes sense. I was worried because of the 0-seconds, but looks like Netlify's just clever.

jennybc:
The full log certainly looks like it going through all the motions and there simply isn't any new content.

https://app.netlify.com/sites/tidyverse-org/deploys/5dccc34d82012772b5cd7743

batpigandme:
Yep. Makes sense. I think I misunderstood what was included in the timings for the GitHub Actions log (trigger took a second, and completing took no time — the Netlify piece isn't in there)
image

PR #375, opened 2019-12-11
is vroom part of tidyverse?

wibeasley:


batpigandme:
Nope it's in r-lib https://vroom.r-lib.org/

batpigandme:
Thanks, though.

hadley:
And vroom is going to get integrated in readr in the not too distant future.

wibeasley:
oh, that's good to know.  Thanks for the heads up.  I've been writing a lot of vroom code recently when readr would have worked just as well.  I'll go back to readr (for the small datasets) so I don't need to update the code later.

Issue #377, opened 2020-01-07
Problems fetching blog posts using a feed reader


hplieninger:
I would like to add https://www.tidyverse.org/blog/ to a feed reader app. However, the file https://www.tidyverse.org/blog/index.xml seems problematic for the feed readers I tried (e.g., Thunderbird, Feedbro), probably because of the date format. For example, the date seems to be `12 Mar 127 120:00 +0000` for the hardhat post.

Apart from that, I would love an RSS button/link just like that on, for example, https://blog.rstudio.com/.

I found this place more appropriate than https://community.rstudio.com/ for my problem; I hope you agree, and I apologize if you disagree.

batpigandme:
Thanks, @hplieninger, this is definitely the right spot! 

@apreshill, do you know if this is something we should change on the theme level, or is this specific to the tidyverse blog?

batpigandme:
Notes to self:
* Pre-migration I had changed rss.xml to use a specific format (RFC 2822) in order to work with a feed aggregator RStudio is using on its site https://github.com/tidyverse/tidyverse.org/pull/168/files
* For the same reason, I added another rss template to the theme we had been using before: https://github.com/tidyverse/tidyverse.org/commit/924a2f0fb1527c1516f6a02a9e9e69a357bb9f2d
* It looks like tourmaline uses RFC 2822 for rss.xml
https://github.com/tidyverse/tidyverse.org/blob/44654367b03572203d7e56e6238970a39a305e26/themes/hugo-tourmaline/layouts/_default/rss.xml#L11
* See also Hugo's [rss template lookup order](https://gohugo.io/templates/rss/#the-embedded-rss-xml)

batpigandme:
Closed by #378

PR #378, opened 2020-01-07
Add section-level RSS template

batpigandme:
* Adds a site-specific template for an RSS feed (see #377 — does not fix in full yet, since we haven't added a button)

cc @apreshill


hadley:
I have no idea whether this is correct or not, so you should just merge if you think it fixes the probelm.

batpigandme:
Requested a review from @apreshill now that she's on the repo

apreshill:
LGTM! After conf, I can re-factor the base theme to enable this too- thanks for figuring it out!

PR #381, opened 2020-01-15
add themis 0.1.0 post

EmilHvitfeldt:
I didn't add any Acknowledgements section as I was the sole contributor code-wise for this package. @topepo could be added for helpful comments and guidance if it seems warranted.


content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,105 @@
...
+
+We're chuffed to announce the release of [themis](https://github.com/tidymodels/themis) is now on CRAN. [themis](https://tidymodels.github.io/themis/) implements a collection of new steps for the [recipes](https://github.com/tidymodels/recipes) package to deal with unbalanced data. themis is still in early development so any and all feedback is highly appreciated.
+
+```{r, message=FALSE}
+library(modeldata)
+library(recipes)
+library(themis)
+```
+
+In a classification context, a dataset is said to be unbalanced if there is an unequal number of observations in each class. Many models perform best when the number of observations is equal and will thus have a tendency to struggle with unbalanced data.
```
batpigandme:
```suggestion
In a classification context, a dataset is said to be unbalanced if there is an unequal number of observations in each class. Many models perform best when the number of observations is equal and, thus, tend to struggle with unbalanced data.
```

batpigandme: I think it's easiest to just stick with present tense there. 🙂

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,105 @@
...
+
+The steps in this package can be divided into two camps:
+
+- Ones that remove observations from the majority class(es) and
```
batpigandme:
```suggestion
- Ones that remove observations from the majority class(es), and
```

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,105 @@
...
+- Ones that add observations to the minority class(es).
+
+One is not restricted to do only one action and is thus able to mix and match by for example first removing observations from the majority class followed by adding observations to the minority class to achieve the balance you want.
```
batpigandme: By "One" here you're referring to the reader or the one doing the modelling, right? English is sorely lacking in the indefinite-pronoun department (meaning "one"), and I think it's trickier here because you've just used "Ones" in the list above to refer to types of steps. Thus, I would fully switch to using "You" here (e.g. _"You can do more than one action, and, thus, are able to mix and match by, for example, removing observations..."_)

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,105 @@
...
+
+## Over-sampling steps
+
+Over-sampling is adding observations to the minority class. Currently `step_upsample()`, `step_smote()`, `step_bsmote()`, `step_adasyn()` and `step_rose()` are available. They have slightly different requirements according to the data they can handle, most needs all numeric with no missing data, but those requirements can and should be handled by previous steps.
```
batpigandme:
```suggestion
Over-sampling involves adding observations to the minority class. Currently `step_upsample()`, `step_smote()`, `step_bsmote()`, `step_adasyn()`, and `step_rose()` are available. They have slightly different requirements according to the data they can handle; most need all numeric with no missing data, but those requirements can and should be handled by previous steps.
```

batpigandme: We're an Oxford comma kind of crowd 😏

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,105 @@
...
+
+In the following example, let's look at the `okc` dataset. and we can see that the imbalance is 1-to-6.
+
+```{r}
+data("okc")
+
+table(okc$Class)
+```
+
+We will use `age`, `diet` and `height` in modeling to predict `Class`. Since `diet` is a factor we first need to dummify it before we normalize and perform mean imputation to handle all the missing data.
```
batpigandme:
```suggestion
We will use `age`, `diet` and `height` in modeling to predict `Class`. Since `diet` is a factor, we first need to dummify it before we normalize and perform mean imputation to handle all the missing data.
```

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,105 @@
...
+  step_smote(Class) 
+
+rec %>%
+  prep() %>%
+  juice() %>%
+  pull(Class) %>%
+  table()
+```
+
+And we see that the resulting dataset has a perfectly even distribution. All the over-sampling steps share the parameter `over_ratio` which specifies the desired ratio between the biggest class and the smallest class. It defaults to 1 for an even distribution but can be set to something like `0.5` to have the minority class become half the size of the majority class
```
batpigandme:
```suggestion
And we see that the resulting dataset has a perfectly even distribution. All the over-sampling steps share the parameter `over_ratio`, which specifies the desired ratio between the biggest class and the smallest class. It defaults to 1 for an even distribution but can be set to something like `0.5` to have the minority class become half the size of the majority class.
```

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,105 @@
...
+rec %>%
+  prep() %>%
+  juice() %>%
+  pull(Class) %>%
+  table()
+```
+
+## Under-sampling steps
+
+Under-sampling is removing observations from the majority class. Currently `step_downsample()`, `step_nearmiss()` and `step_tomek()` are available. These steps should have the same user experience as the previous steps as they have a similar shared parameter `under_ratio` which is the ratio between the smallest and the biggest class. Simply using `step_downsample()` removes samples in such a case that all classes as the size of the smallest class.
```
batpigandme:
```suggestion
Under-sampling is removing observations from the majority class. Currently `step_downsample()`, `step_nearmiss()`, and `step_tomek()` are available. These steps should have the same user experience as the previous steps as they have a similar shared parameter `under_ratio`, which is the ratio between the smallest and the biggest class. Simply using `step_downsample()` removes samples in such a way that all classes as the size of the smallest class.
```

batpigandme: I think you're missing a word or two at the end of that last sentence "that all classes **missing word** as the size of the smallest". I didn't want to put one in for you, since I don't know what the most accurate term would be.

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,105 @@
...
+  step_normalize(all_predictors()) %>%
+  step_meanimpute(all_predictors()) %>%
+  step_downsample(Class) 
+
+rec %>%
+  prep() %>%
+  juice() %>%
+  pull(Class) %>%
+  table()
+```
```
batpigandme: I think @EmilHvitfeldt has a separate PR adding acknowledgements here, but I recommend making the edit in this PR.

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
EmilHvitfeldt: I was indeed trying to use an indefinite pronoun, will switch to "you"

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
EmilHvitfeldt: I did a whole/half-sentence rewrite of this sentence

EmilHvitfeldt:
Thanks for all feedback

- deleted .html.
- rendered with `blogdown::serve_site()`
- resolved all changes
- added a note on the presence of references in help pages
- changed "Over-sampling" to "Hybrid-sampling" 
- Mentioned difference between upsampling and synthetic data point generation

content/blog/2020-01-themis-0-1-0/index.Rmarkdown
```diff
@@ -0,0 +1,104 @@
+---
+title: themis 0.1.0
+author: Emil Hvitfeldt
+date: '2020-01-13'
+slug: themis-0-1-0
+description: 
+  themis 0.1.0 is now available on CRAN. Provides additional steps for `recipes`
```
batpigandme:
```suggestion
  themis 0.1.0 is now available on CRAN. Provides additional steps for recipes
```

batpigandme: Sorry, I lied, one more change: we don't use backticks around package names. Just caught this one when looking at the actual preview.

EmilHvitfeldt:
The date and description changed!

Issue #384, opened 2020-02-01
Reflect change from retired to superseded 


EmilHvitfeldt:
https://www.tidyverse.org/lifecycle/

batpigandme:
There's some nuance here, which is better suited to the lifecycle package. superseded is for functions, as opposed to packages, I believe.

jennybc:
I'd like a ruling on this too. This relates to https://github.com/r-lib/usethis/pull/1004 which changes "retired" to "superseded" in the lifecycle-related functions of usethis.

hadley:
I think we should use superseded for both functions and packages.

hadley:
Somewhat related, I am also wondering if maturing only makes sense for packages, not functions.

Issue #274, opened 2019-02-13
Lifecycle-badge definitions as they apply to functions


maurolepore:
I see that lifecycle badges are increasingly used not only to refer to packages but also to functions within packages. But the definition of each stage remains focused on packages. 

The solution may be as simple as replacing 'package' with 'package/function'. E.g. 

> We are no longer convinced that a questioning package/function is the optimal approach, but we don’t yet know what a better approach is.

batpigandme:
These now exist in the lifecycle package. (See [Getting started here](https://lifecycle.r-lib.org/articles/lifecycle.html)). I would rather _not_ maintain the definitions in two separate places, as that means we'd have to remember to update the tidyverse.org page if we change the wording in the lifecycle docs. But adding a link to the lifecycle definitions in the tidyverse page seems reasonable. (No need to PR, @maurolepore, as I'm working through the entire site over the next few weeks).

hadley:
We might also consider adding a package lifecycle vignette to the lifecycle package so we can link both resources to the same place.

jennybc:
I'm staring at lifecycle stuff in usethis and believe we have now arrived at a situation where the lifecycle stages at <https://www.tidyverse.org/lifecycle/> disagree with those at <https://lifecycle.r-lib.org/articles/lifecycle.html> (the "getting started" vignette of the lifecycle package). I'm going to align with the lifecycle package (?).

Agree:

* experimental
* maturing
* stable
* questioning
* superseded

Disagree:
* archived (only tidyverse.org)
* dormant (only tidyverse.org)
* soft-deprecated (only lifecycle)
* deprecated (only lifecycle)
* defunct (only lifecycle)

batpigandme:
FYI, if you're tying the diagram in in any way, I just opened a PR in lifecycle so that the box colours match up with the badges: https://github.com/r-lib/lifecycle/pull/53#issue-457230407

PR #393, opened 2020-03-12
Update lifecycle to use superseded

batpigandme:
This is a minimal fix to #384 (read: I think we might want to expand the language eventually).

* Add new source (graffle) file for lifecycle diagram.
* Add updated lifecycle diagrams as svg and png.
* Change retired section to "superseded".


batpigandme:
> We also need to update this page to talk about functions as well as packages

I think that's issue #274

PR #398, opened 2020-03-25
pkgdown 1.5.0

hadley:
@jayhesselberth can you take a look too please?

TO DO:

* [ ] remove `/dev` once site re-published


hadley:
@batpigandme I wanted to just get this out the door, so no need to review unless you see something particularly horrific.

Issue #406, opened 2020-04-06
Figure out how to avoid re-rendering old posts


batpigandme:
Currently, old posts that haven't been changed are getting re-rendered when contributors try to render the site (i.e. run `blogdown::serve_site()`). We want to avoid this (for obvious reasons). cc @apreshill

hadley:
I think this is a fundamental flaw with the design of blogdown/hugo, and I suspect @jennybc's hack at  https://github.com/tidyverse/tidyverse.org#troubleshooting is the best we can do.

hadley:
Solved by conversion to hugodown.

PR #407, opened 2020-04-07
WIP: Modify last line of rendered blog post files

batpigandme:
Adds a blank space before the end of the last line in all rendered post files.

Could've used an "OR", I'm sure, but did this with:
```sh
# Working from tidyverse.org/content/blog directory
find . -name "index.html" | xargs -L1 sed -i '' '$s/.*/ &/'

find . -name "index.md" | xargs -L1 sed -i '' '$s/.*/ &/'

find . -name "index.markdown" | xargs -L1 sed -i '' '$s/.*/ &/'
```
For `.html` files this is just a blank space before the closing `</div>`, and `.markdown` and `.md` ignore a leading single space.

This should allow us to use git commit as mtime to avoid rebuilding old posts (since, all the source and rendered files were in the same commit from the bundling migration). 

One piece of #406.


apreshill:
Hi @batpigandme - would it help to add a line break after the `{{ .Content }}` to the Hugo code for all the single page layouts for future posts?

jennybc:
> This should allow us to use git commit as mtime to avoid rebuilding old posts (since, all the source and rendered files were in the same commit from the bundling migration)

I don't think this is going to help. Because the problem will just continue to be recreated as we write more posts. Whenever any individual pulls after a period of not working on the blog, they'll get both source and rendered product for all posts added in the interim. And we're back where we started.

batpigandme:
OK, maybe I misunderstood, I thought that from what Jim was saying (and one of the options in the Hugo serve docs) it'd be possible to use git commits as mtimes. 

I'm gonna look in a couple different spots for avoiding the problem moving forward-- obviously Hugo's fast serve logic isn't doing it for us. 

@apreshill, I don't actually want the extra space for anything, it was just a minimal committable change!

jennybc:
Maybe I misunderstand? But it seems like most people will commit their post's (source, rendered) file pair in the same commit? So they will always end up with the same mtime, because the have the same SHA.

batpigandme:
No, you're right about that part Jenny! That's where I'm digging for another approach. Hopefully this PR will end up being irrelevant.

hadley:
Closing, because I'm pretty sure the problem lies with blogdown and there's nothing we can easily do to fix it (except to regularly re-run the fix in the readme)

PR #409, opened 2020-04-10
dplyr 1.0.0: rowwise() features

hadley:


content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+
+Note that the development version won't become 1.0.0 until it's released, but it has all the same features.
+
+```{r setup}
+library(dplyr, warn.conflicts = FALSE)
+```
+
+## Basic operation
+
+`rowwise()` works like `group_by()` in the sense that it doesn't change what the data looks like; it changes how dplyr verbs operate on it. Let's see how this works with a simple example. Here I have some imaginary test results for students in a class:
```
batpigandme, line 44:
```suggestion
`rowwise()` works like `group_by()` in the sense that it doesn't change what the data looks like; it changes how dplyr verbs operate on the data. Let's see how this works with a simple example. Here I have some imaginary test results for students in a class:
```

batpigandme: I don't love my wording change, but the double "it" in that last part was a tad confusing.

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+  student_id = 1:4, 
+  test1 = 10:13, 
+  test2 = 20:23, 
+  test3 = 30:33, 
+  test4 = 40:43
+)
+df
+```
+
+It'd like to be able to compute the mean of the test scores for each student, but `mutate()` and `mean()` do what I want:
```
batpigandme, line 57:
```suggestion
I'd like to be able to compute the mean of the test scores for each student, but `mutate()` and `mean()` don't do what I want:
```

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+
+```{r}
+df %>% mutate(avg = mean(c(test1, test2, test3, test4)))
+```
+
+The problem is that I'm getting a mean over the whole data frame, not for each student. I can resolve this problem a get a mean for each student by creating create a "row-wise" data frame with `rowwise()`:
```
batpigandme, line 63:
```suggestion
The problem is that I'm getting a mean over the whole data frame, not for each student. I can resolve this problem of getting a mean for each student by creating a "row-wise" data frame with `rowwise()`:
```

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+```{r}
+rf
+```
+
+`rf` _looks_ very similar to `df`, but _behaves_ very differently:
+
+```{r}
+rf %>% mutate(avg = mean(c(test1, test2, test3, test4)))
+```
+An additional advantage of `rowwise()` is that it's paired with `c_across()`, which works like `c()` but uses the same tidyselect syntax as `across()`. That makes it easy to operate on multiple variables:
```
batpigandme, line 80:
```suggestion
An additional advantage of `rowwise()` is that it's paired with [`c_across()`](https://dplyr.tidyverse.org/dev/reference/across.html), which works like `c()` but uses the same tidyselect syntax as `across()`. That makes it easy to operate on multiple variables:
```

batpigandme: I know we don't consistently link to the function references, but I think it's handy here.

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+
+```{r}
+rf %>% mutate(avg = mean(c_across(starts_with("test"))))
+```
+### Other ways of achieving the same result
+
+Some summary functions have alternative ways to compute row-wise summaries that take advantage of built-in vectorisation. For example, if you wanted to compute the sum, you could use `+`:
```
batpigandme, line 87:
```suggestion
Some summary functions have alternative ways of computing row-wise summaries that take advantage of built-in vectorisation. For example, if you wanted to compute the sum, you could use `+`:
```

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+)
+```
+
+Where these functions exist, they'll usually be faster than `rowwise()`. The advantage of `rowwise()` is that it works with any function, not just those that are already vectorised.
+
+## List-columns
+
+`rowwise()` is useful for computing simple summaries, but its real power comes when you use it with list-columns. Because lists can contain anything, you can use list-columns to keep related objects together, regardless of what type of thing they are. List-columns give you a convenient storage mechanism and `rowwise()` gives you a convenient computation mechanism.
+
+Let's make those ideas concrete by creating a data frame a list-column. A little later, we'll come back to how you might actually get a list-column in a more realistic situation. The following data frame uses list columns to store things that would otherwise be challenging:
```
batpigandme, line 114:
```suggestion
Let's make those ideas concrete by creating a data frame with a list-column. A little later, we'll come back to how you might actually get a list-column in a more realistic situation. The following data frame uses list columns to store things that would otherwise be challenging:
```

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+df %>% 
+  rowwise() %>% 
+  summarise(
+    x_length = length(x),
+    y_type = typeof(y),
+    z_call = z(1:5)
+  )
+```
+
+This makes a row-wise `mutate()` or `summarise()` a general vectorisation tool, in the same way as the apply family in base R or the map family in purrr. It's now much simpler to solve a number of problems where we previously recommended learning about `map()`, `map2()`, `pmap()` and friends.
```
batpigandme, line 141:
```suggestion
This makes a row-wise `mutate()` or `summarise()` a general vectorisation tool, in the same way as the apply family in base R or the map family in purrr do. It's now much simpler to solve a number of problems where we previously recommended learning about `map()`, `map2()`, `pmap()` and friends.
```

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+df %>% mutate(avg = mean(c(test1, test2, test3, test4)))
+```
+
+The problem is that I'm getting a mean over the whole data frame, not for each student. I can resolve this problem a get a mean for each student by creating create a "row-wise" data frame with `rowwise()`:
+
+```{r}
+rf <- rowwise(df, student_id)
+```
+
+`rowwise()` doesn't need any additional arguments unless you have variables that identify the rows, like `student_id` here. Much like grouping variables, they will be automatically preserved when you `summarise()` the data.
```
batpigandme, line 69:
```suggestion
`rowwise()` doesn't need any additional arguments unless you have variables that identify the rows, like `student_id` here. Much like grouping variables, identifier variables will be automatically preserved when you `summarise()` the data.
```

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+
+```{r}
+df %>%
+  rowwise(id) %>%
+  summarise(x = runif(n, min, max))
+```
+
+Note that `id` is preserved in the output here because we defined it as an identifier variable in the call to `rowwise()`.
+
+[`vignette("rowwise")`](https://dplyr.tidyverse.org/dev/articles/rowwise.html#repeated-function-calls-1) expands on this idea to show you can generate parameter grids and vary the random distribution used in each row.
```
batpigandme, line 178:
```suggestion
[`vignette("rowwise")`](https://dplyr.tidyverse.org/dev/articles/rowwise.html#repeated-function-calls-1) expands on this idea to show how you can generate parameter grids and vary the random distribution used in each row.
```

content/blog/2020-04-dplyr-1-0-0-rowwise/index.Rmarkdown
```diff
@@ -0,0 +1,205 @@
...
+
+### Group-wise models
+
+The new `nest_by()` function works similarly to `group_by()` but instead of storing the grouping data as metadata, visibly changes the structure. Now we have three rows (one for each group), and we have a list-col, `data`, that stores the data for that group. Also note that the output is a `rowwise()`; this is important because it’s going to make working with that list of data frames much easier.
```
batpigandme, line 182:
```suggestion
The new `nest_by()` function works similarly to `group_by()` but instead of storing the grouping data as metadata, visibly changes the structure. Now we have three rows (one for each group), and we have a list-col, `data`, that stores the data for that group. Also note that the output is a `rowwise()` object; this is important because it’s going to make working with that list of data frames much easier.
```

batpigandme: You can obviously use a different word, but I think what you're going for here is the rowwise equivalent of "a grouped data frame," so you need _some_ noun (unless you take out the a, and say "the output is `rowwise()`", in which case I don't think rowwise should be formatted as a function).

hadley: Yeah, sounds good.

PR #411, opened 2020-04-15
Set aria-hidden for images on blog page

### LINKED ISSUE #205: Set aria-hidden on image links on articles page
ie. `"aria-hidden="true"` so that there aren't two identical links next to each other

batpigandme:
* Fixes #205
* Edit to theme

### DIFF
--- a/themes/hugo-tourmaline/layouts/blog/list.html
+++ b/themes/hugo-tourmaline/layouts/blog/list.html
@@ -13,7 +13,7 @@
           <div class="listItem">
              {{ $card := (.Resources.ByType "image").GetMatch "*-sq*" }}
              {{ with $card }}
-              <a class="itemImage project" href="{{ $post.RelPermalink }}" style="background-image: url('{{ .RelPermalink }}');"></a>
+              <a class="itemImage project" href="{{ $post.RelPermalink }}" aria-hidden="true" style="background-image: url('{{ .RelPermalink }}');"></a>
               {{ end }}
             <div class="itemDetails">
               <div class="itemHeader">
--- a/themes/hugo-tourmaline/layouts/blog/list.html
+++ b/themes/hugo-tourmaline/layouts/blog/list.html
@@ -13,7 +13,7 @@
           <div class="listItem">
              {{ $card := (.Resources.ByType "image").GetMatch "*-sq*" }}
              {{ with $card }}
-              <a class="itemImage project" href="{{ $post.RelPermalink }}" aria-hidden="true" style="background-image: url('{{ .RelPermalink }}');"></a>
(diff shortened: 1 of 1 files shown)


batpigandme:
@apreshill So, I set the blog list to have the images as `aria-hidden = TRUE`, but I still get [accessibility warnings](https://wave.webaim.org/report#/https://deploy-preview-411--tidyverse-org.netlify.com/blog/) because there's no text (since it's just an image), and the link is redundant with the title in the item details (which is the text that's "missing" from the images).

Is there a reason to have the images also be links? If we just had them as background images that weren't links, we wouldn't have the link duplication _or_ the warning re. a link without a title.

batpigandme:
OK - shoot, I'm an idiot, obviously making the change in the same branch made the change in the deploy preview and, thus, in the accessibility link, _but_, as you can see there, un-linking the images means there's no accessibility error there.

apreshill:
@mara Would it be possible to give me special Netlify privileges for this site? I don't have access to the deploy preview links.

batpigandme:
Ah, let me see. I vaguely remember there being some weird thing where it was like a per-person subscription model (@topepo asked about this a while back, I think), but I'll look into it right now.

batpigandme:
OK, yeah! We're on the pro plan, which has 3 seats (Hadley, Jenny, and I occupy those right now). Does RStudio have a netlify plan we could piggyback on? @hadley, any thoughts?

batpigandme:
>  I don't have access to the deploy preview links.

Also, here's the deploy preview link: https://deploy-preview-411--tidyverse-org.netlify.app/blog/

hadley:
Currently we're on the $45/month plan; to get more seats it appears that we need to upgrade to the enterprise plan which would cost $1000 a month 😱.

But the deploy preview shouldn't require any auth?

batpigandme:
Yeah, I, too was surprise by that… Definitely doesn't seem like it's worth the bump in price

themes/hugo-tourmaline/layouts/blog/list.html
```diff
@@ -13,7 +13,7 @@
           <div class="listItem">
              {{ $card := (.Resources.ByType "image").GetMatch "*-sq*" }}
              {{ with $card }}
-              <a class="itemImage project" href="{{ $post.RelPermalink }}" style="background-image: url('{{ .RelPermalink }}');"></a>
+              <div class="itemImage project" aria-hidden="true" style="background-image: url('{{ .RelPermalink }}');"></div>
```
hadley, line 16: Would it be marginally cleaner to make this an `` with `alt = ""`?

apreshill:
I believe this PR can be closed now, with the latest Hugo theme upgrade
Screen Shot 2020-07-17 at 3 04 54 PM

PR #418, opened 2020-04-18
Disable highlightjs

apreshill:
Now that pygments can be used in Hugo, I noticed that our previous highlightjs was causing a conflict and overwriting some of the highlight elements. It was also leading to what you see when you currently load a post with code in it- the pygments version flickers but then turns gray, because the highlightjs kicks in. The Hugo theme now has a site parameter to disable highlightjs sitewide.

Before: 
Screen Shot 2020-04-17 at 8 07 33 PM

After: 
Screen Shot 2020-04-17 at 8 07 44 PM

### DIFF
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -16,6 +16,7 @@ hex_base_url = ".tidyverse.org"
 mainSections = ["blog"]
 
 # options for highlight.js (version, additional languages, and theme)
+disable_highlightjs = "true"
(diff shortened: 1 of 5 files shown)


apreshill:
I'll let someone merge in who can check the Netlify deploy preview. It all looks fine locally but 😬

batpigandme:
Looks like this takes care of the low-contrast accessibility issue, too! Thanks, Alison.

PR #427, opened 2020-04-28
Self cleaning card

apreshill:
The twitter card for @jennybc's recent post wouldn't render because of file size limitations. I shrunk it down and it appears to work now.


apreshill:
It works :tada: (+ thank you @batpigandme - I now see where the deploy previews show up under checks)
Screen Shot 2020-04-28 at 9 16 15 AM

jennybc:
Thanks!

This would be good to address structurally somehow. At the least, diagnose? Even better, fix. But that's no reason to not fix this one.

PR #435, opened 2020-05-11
Use `where()` in dplyr blog posts

lionel-:


content/blog/2020-04-dplyr-1-0-0-colwise/index.Rmarkdown
```diff
@@ -24,6 +24,19 @@ This post is the latest in a series of post leading up the the dplyr 1.0.0 relea
 
 Today, I wanted to talk a little bit about the new `across()` function that makes it easy to perform the same operation on multiple columns.
 
+
+### _Update notice_
```
hadley, line 28: I think you could omit the update notice here, since the select syntax isn't the primary point of the article (and we're just making the change so that people coming to the post in the future aren't confused)

lionel-: Just to be clear, I remove it from the across post and leave it in the select post? Or remove it from both?

hadley: Remove it from this post, but leave it in the others.

PR #439, opened 2020-05-16
Convert config to yaml

hadley:


hadley:
Can these be inlined into the main config?

The major advantages to using yaml is that we can put hugodown metadata in this file and I can actually write it without having to lookup the syntax every time.

apreshill:
We can The config directory was one of the modernizations I introduced when making the theme into a proper Hugo theme: https://gohugo.io/getting-started/configuration/

I believe the config directory will make it easier to have multiple language sites, like `tidyverse.org/en` and `tidyverse.org/es` if that is ever of interest.

I'm not sure how long this conversion would take me on the layout-side. It is possible it is quick, or there might be multiple Hugo rabbit holes I need to debug. I'm just not sure!

hadley:
But maybe we can just leave those files as is, and only change this file? I'm going to merge this for now so I can continue hugodown explorations, but we can definitely change back later if we decide it's better to stick with toml.

PR #444, opened 2020-05-25
Move old posts into subdirectories

hadley:


hadley:
Closing in favour of doing with code so can more easily be re-applied.

PR #445, opened 2020-05-25
Remove old articles directory

hadley:


batpigandme:
Pretty sure this is going to break images in old posts 😬

PR #450, opened 2020-06-02
Use styles for syntax highlighting

hadley:
(This is mostly going to be important for new posts, since downlit will perform the syntax highlighting)

FYI @apreshill & @batpigandme


apreshill:
When we update the theme, the CSS file will get overridden. So we could either:

1. move it upstream to https://github.com/rstudio/hugo-tourmaline (probably preferable as we'll be transitioning to hugodown on the other 3 sites) or...

1. add it to your site-specific css: https://github.com/tidyverse/tidyverse.org/blob/master/static/css/tidyverse.css

hadley:
I'm dumping it in the theme css, in the assumption it will get included upstream.

PR #452, opened 2020-06-09
Simplify netlify.toml

hadley:
* Set HUGO_VERSION for all builds

* Set command for all builds

* Rely on `hugo` defaulting to production environment, as [documented](https://gohugo.io/getting-started/configuration/#configuration-directory)

* Removed `HUGO_ENABLEGITINFO` since, as far as I can tell, it's not documented anywhere, and comes only from people copy and pasting from <https://gohugo.io/hosting-and-deployment/hosting-on-netlify/>

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,20 +1,6 @@
 [build]
   publish = "public"
-  command = "hugo"
-
-[context.production.environment]
-  HUGO_VERSION = "0.68.3"
-  HUGO_ENV = "production"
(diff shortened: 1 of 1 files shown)


apreshill:
This will build & publish all posts with a future publish date (i.e., `publishdate: "2020-08-20T00:00:00Z"` or `date: "2020-08-20T00:00:00Z"`). See:

https://flamboyant-fermat-456014.netlify.app/#posts

PR #459, opened 2020-07-13
Fix code font sizing in headers + TOC

apreshill:
This fix makes code font size for inline code relative to direct or nearest parent (`em`), not root font size (`rem`). Pulls from upstream fix in Hugo Graphite theme.

Before:
https://www.tidyverse.org/blog/2020/07/tune-0-1-1/#better-autoplot

After:
https://deploy-preview-459--tidyverse-org.netlify.app/blog/2020/07/tune-0-1-1/#better-autoplot


hadley:
Thanks!

Issue #212, opened 2018-08-30
Update hex logos image on homepage


mine-cetinkaya-rundel:
The current image on the homepage shows hex logos 6 packages, I think it should show all 8 in core tidyverse.

This would have the additional benefit of getting more 🐱 on the homepage. 🎉

hadley:
Duplicate of #95

hadley:
Looks like this is now controlled by `layout/index.html`.

@apreshill in principle, it looks like I could just add a `col1` section. Is it ok for me to do that in `layout/index.html` or does it need to do upstream in a theme file somewhere?

apreshill:
Hi, 

Here is what would need to change. It could change just in tidyverse.org, but if we go to all the trouble of changing the CSS it would be great to do it in the theme so tidymodels.org could also use it easily.

1. The current rows/columns are defined in the CSS, done I believe by Robby S. We would need new CSS classes for any new hex positions: https://github.com/rstudio/hugo-tourmaline/blob/master/static/css/main-site.css#L1205-L1232

1. We would also need to consider how this will look on mobile: https://github.com/rstudio/hugo-tourmaline/blob/master/static/css/main-site.css#L1234-L1240

1. Once that is done, I parameterized this so you can add them to your `index.html` yaml and the corresponding layout file will need to have new slots added (this is pretty quick!): https://github.com/rstudio/hugo-tourmaline/blob/master/layouts/index.html#L11-L25

hadley:
Re-opening this issue since I think I closed the wrong one.

juliasilge:
With us finally getting a hex for workflows, this would be nice to figure out for tidymodels.org as well.

Issue #461, opened 2020-07-24
Update logos on https://www.tidyverse.org/packages/#core-tidyverse


courtiol:
While the main page of the website links to up to date versions of the hex-logos for dplyr and tidyr, the page https://www.tidyverse.org/packages/#core-tidyverse does not.

PR #464, opened 2020-07-30
Moar hexes

apreshill:
This PR is a minor Hugo graphite theme update. It adds:

+ 2 more hexes to the homepage (in band 1)
+ alt text to the cover image (in band 2)
+ a `packages.yml` that now populates the packages page via a shortcode (so we don't need to edit raw html with deeply nested divs)
+ fixes the issue @lionel- ran into with the partial meta layout file with no images in the post bundle
+ minor CSS updates and layout fixes that came up in testing other sites


hadley:
Does this fix #212?

hadley:
BTW it'd be better to not hot-linking to the logos on hex-stickers because they're high-res versions that are often 2+ MB! 😱

apreshill:
Hi @hadley - this does fix #212 and #461. I updated to make all hex images self hosted now, as opposed to hot linked, on the homepage and the packages page.

hadley:
Looks awesome, thanks!

PR #491, opened 2020-12-09
update RSS layout to fix image rendering

apreshill:
Please wait to merge in until we can test deploy preview with RSS reader, thanks!


themes/hugo-graphite/layouts/_default/rss.xml
```diff
@@ -32,7 +32,9 @@
       <link>{{ .Permalink }}</link>
       <pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
       <guid>{{ .Permalink }}</guid>
-      <description>{{ .Content | html }}</description>
+      {{- $content := replaceRE "a href=\"(#.*?)\"" (printf "%s%s%s" "a href=\"" .Permalink "$1\"") .Content -}}
```
hadley, line 35: Could you please use a comment here to link back to that blog post that explains how this works? 

It's unfortunate that hugo doesn't provide something built in for this case, because the regexp approach is going to be fragile (i.e. it relies on the alt attribute coming after the src attribute).

hadley (approved):

batpigandme (approved):

PR #503, opened 2021-02-27
clock 0.1.0

DavisVaughan:
- [ ] Wait for clock to hit CRAN
- [ ] Update blog post date


DavisVaughan:
@batpigandme, clock isn't on CRAN yet, and probably won't be for another 2 weeks or so, but if you have any time to review the release post in the meantime, that would be awesome! It is large, but I wanted to make a motivating case for using clock.

DavisVaughan:
Also, @hadley if you have any thoughts about the lubridate section near the beginning of the article, I'd welcome them!

jennybc:
Unsolicited feedback: the post (and the package) looks great!

I came away with 2 meta comments:

---

I think you bury the lede a bit in the intro. I would consider pulling a short statement about "reducing the agony caused by time zones" into the first couple of sentences, i.e. before the first set of bullet points. I realize it comes up in the first bullet, but behind some rather technical terminology ("orthogonal date time classes"). I think there are lots of people who aren't very clear on what their date time problems have been in the past, nor on how to solve them. But they know for damn sure that time zones were the root cause.

---

About the general issue of invalid or ambiguous dates or date times. I feel like fewer people will immediately recognize this as a problem they have (although many people *do* have this problem and just aren't aware of it (yet)!). A relevant point here seems to be that when you develop code with example data, you might not stumble across any of these. But as soon as your thing goes into production or gets used on a new or larger dataset, that's when you will learn about this the hard way.

Has anyone ever quantified this risk for people? I.e. if you add 1 month to every date in the year, what proportion of those dates are invalid? In this particular case, it would be easy for you to work out. I'm just thinking about a factoid that helps people grasp these problems are rare enough that it's easy to forget about them, but common enough that they will eventually bite you.

content/blog/clock-0-1-0/index.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  mutate(
+    date2 = date + months(1),
+    date3 = date %m+% months(1),
+    date4 = add_with_rollback(date, months(1), roll_to_first = TRUE)
+  )
+```
+
+With Dates, the hardest part about `%m+%` is remembering to use it. It is a common bug to forget to use this helper until *after* you have been bitten by an invalid date issue with an unexpected `NA`.
+
+The overarching goal of clock is to protect you, the hard-working data scientist who shouldn't have to be an expert in date-times, from issues like this by erroring early and often, rather than letting this slip through unnoticed, only to cause hard to debug issues down the line.
```
hadley, line 201: I think you could ratchet this down a notch

content/blog/clock-0-1-0/index.Rmd
```diff
@@ -0,0 +1,492 @@
...
+flights_hm %>%
+  select(date) %>%
+  mutate(
+    datetime = as.POSIXct(date),
+    datetime_utc = date_set_zone(datetime, "UTC")
+  ) %>%
+  head(n = 3)
+```
+
+To get what we want, we need to convince the `date` column to "forget" that it is UTC, then add on the America/New_York time zone. We'll do this by going through a new intermediate type called naive-time, a date-time type with a yet-to-be-specified time zone. We'll talk more about naive-times in the Time Points section below. This retains the printed time as we expected.
```
hadley, line 247:
```suggestion
To get what we want, we need to convince the `date` column to "forget" that it is UTC, then add on the America/New_York time zone. With clock, we'll do this by going through a new intermediate type called naive-time, a date-time type with a yet-to-be-specified time zone. We'll talk more about naive-times in the Time Points section below. This retains the printed time as we expected.
```

hadley: I think you could sell this a bit more as one of the big advantages of clock

content/blog/clock-0-1-0/index.Rmd
```diff
@@ -0,0 +1,492 @@
...
+
+# 02:23:00 -> NA
+add_days(problem, 2, nonexistent = "NA")
+```
+
+I recommend `"roll-forward"` or `"roll-backward"`, as these retain the _relative ordering_ of `datetime`, an issue that you can read about [here](https://r-lib.github.io/clock/articles/articles/motivations.html#nonexistent-time-1).
+
+Unlike with invalid dates, lubridate does not provide any tools for resolving nonexistent times.
+
+There are another class of daylight saving time issues related to _ambiguous times_. You resolve them in a similar way to what was done with nonexistent times. If you're interested, you can read more about those ambiguous times [here](https://r-lib.github.io/clock/articles/articles/motivations.html#ambiguous-time-1).
```
hadley, line 336: I'd provide one more sentence here (i.e. something about being two 2:30 AMs)

content/blog/clock-0-1-0/index.Rmd
```diff
@@ -0,0 +1,492 @@
...
+invalids
+```
+
+The `ymd2` column directly contains the invalid dates, `2013-02-29` and `2013-02-30`! You also get a note up in the header, `i=2`, letting you know that you have two invalid dates in that column. You can resolve these dates at any time using `invalid_resolve()`, providing an invalid resolution strategy like we did earlier. Or, you can ignore them if you expect them to be resolved naturally in some other way. For example, if our end goal was to add 1 month, then fix the day of the month to the 15th, then these invalid dates would naturally resolve themselves:
+
+```{r}
+mutate(invalids, ymd3 = set_day(ymd2, 15))
+```
+
+The important thing is that they _eventually_ get resolved. You must resolve invalid dates before converting to another calendar or to a Date / POSIXct.
```
hadley, line 425: Mention function that tells you if a date is valid or not

content/blog/clock-0-1-0/index.Rmd
```diff
@@ -0,0 +1,492 @@
...
+
+```{r, error=TRUE}
+as.Date(invalids$ymd2)
+```
+
+### Time points and Zoned-times
```
hadley, line 431:
```suggestion
### Time points and zoned-times
```

content/blog/clock-0-1-0/index.Rmd
```diff
@@ -0,0 +1,492 @@
...
+
+The daylight saving time section of this post was complicated by the need to work around time zones. If your analysis doesn't actually require time zones, you can represent a date or date-time using a _naive-time_. This date-time type makes no assumption about the current time zone, instead assuming that there is a yet-to-be-specified time zone that hasn't been declared yet. Naive-times vary in precision from day up to nanosecond.
```
hadley, line 433:
```suggestion
The daylight saving time section of this post was complicated by the need to work around time zones. If your analysis doesn't actually require time zones, you can represent a date or date-time using a _naive-time_. This date-time type makes no assumption about the current time zone, instead assuming that there is a yet-to-be-specified time zone that hasn't been declared yet. 
```

content/blog/clock-0-1-0/index.Rmd
```diff
@@ -0,0 +1,492 @@
...
+-   Implements a [high level API](https://r-lib.github.io/clock/reference/index.html#section-high-level-api) for Date and POSIXct classes that lets
+    you get productive quickly without having to learn the details of
+    the underlying classes.
+
+-   Requires explicit handling of invalid dates (e.g. what date is one
+    month after January 31st?) and nonexistent or ambiguous times
+    (caused by daylight saving time issues).
+
+-   Is built on the C++ [date](https://github.com/HowardHinnant/date)
+    library, which provides a correct and high-performance backend.
```
hadley, line 41: I think it would be a worth a sentence or two here roughly summarising the performance of lubridate vs clock.

hadley (approved):

DavisVaughan:
Thanks @hadley and @jennybc!

@jennybc I've added two new commits addressing your meta-thoughts:

- a5118d7 : Mentions the new date-time types in the opening paragraph, and hints at the fact that they help with time zone agony
- 9d886db : Mentions the relative frequency of invalid dates when doing month based arithmetic (7 times a year), and the relative _infrequency_ of nonexistent/ambiguous times as a nice lead into the Production section

A number of times while writing this article, I thought that clock's mantra should be that it helps you "practice safe data science when working with date-times"

PR #516, opened 2021-06-24
Off-label use of ggsave

thomasp85:
This post is missing a header image which I'll create later, but the text is ready for review


content/blog/off-label-uses-in-ggplot2/index.Rmd
```diff
@@ -0,0 +1,100 @@
...
+
+<!--
+TODO:
+* [ ] Look over / edit the post's title in the yaml
+* [ ] Edit (or delete) the description; note this appears in the Twitter card
+* [ ] Pick category and tags (see existing with `hugodown::tidy_show_meta()`)
+* [ ] Find photo & update yaml metadata
+* [ ] Create `thumbnail-sq.jpg`; height and width should be equal
+* [ ] Create `thumbnail-wd.jpg`; width should be >5x height
+* [ ] `hugodown::use_tidy_thumbnails()`
```
hadley, line 29: Did you run this one? I noticed the square thumbnail took a little while to load.

content/blog/off-label-uses-in-ggplot2/index.Rmd
```diff
@@ -0,0 +1,100 @@
...
+* [ ] Add intro sentence, e.g. the standard tagline for the package
+* [ ] `usethis::use_tidy_thanks()`
+-->
+
+ggplot2 v3.3.4 landed on CRAN recently, and while every release of ggplot2 is cause for celebration, this was merely a patch release fixing a large number of bugs and it thus came and went without much fanfare. However, for a couple of users this release brought an unwelcome and surprise change. We feel that this is a great opportunity to talk a bit about some of the topics that Hadley discussed in his [rstudio::global(2021) keynote](https://www.rstudio.com/resources/rstudioglobal-2021/maintaining-the-house-the-tidyverse-built/) where he addresses the nature of breaking changes.
```
hadley, line 34:
```suggestion
ggplot2 v3.3.4 landed on CRAN recently, and while every release of ggplot2 is cause for celebration, this was merely a patch release fixing a large number of bugs and so it came and went without much fanfare. However, for a couple of users this release brought an unwelcome and surprising change. We feel that this is a great opportunity to talk a bit about some of the topics that Hadley discussed in his [rstudio::global(2021) keynote](https://www.rstudio.com/resources/rstudioglobal-2021/maintaining-the-house-the-tidyverse-built/), particularly the nature of breaking changes.
```

content/blog/off-label-uses-in-ggplot2/index.Rmd
```diff
@@ -0,0 +1,100 @@
...
+  ggsave("my_mpg_plot.png")
+```
+
+Now, if this is the first time you've seen `ggsave()` being added to a plot, you are not alone. This certainly caught us by surprise. Prior to v3.3.4, this actually worked (more on that later) but with the recent release running this code will result in the following error:
+
+    Error: Can't add `ggsave("my_mpg_plot.png")` to a ggplot object.
+
+If you were a user that had used this pattern for saving plots it very much felt like we had removed a feature, pulling the rug out from under your script with no warning. However, this use of `ggsave()` had never been advertised in any of the documentation and while it worked, it could not be considered a feature as such.
+
+For completeness, this is the advertised and supported use of `ggsave()`:
```
hadley, line 54: Might be better to start with this — something like "we designed `ggsave()` to be used ...

content/blog/off-label-uses-in-ggplot2/index.Rmd
```diff
@@ -0,0 +1,100 @@
...
+```{r, eval=FALSE}
+ggplot(mpg) + 
+  geom_point(aes(x = displ, y = hwy))
+
+ggsave("my_mpg_plot.png")
+```
+
+## Off-label saving
+
+The issue described above falls under the category of off-label use that Hadley talks about in his keynote. Off-label use of functions comprise of using functions in a way that only work *by accident*, and are thus susceptible to breakage at any point due to changes in the code. Another common word for this is "a hack", but this term can often imply that the user is full aware of the brittle nature of the setup. Off-label use can just as well be passed on between users to a point where some thinks that this is the correct, supported, way of doing things (this was certainly the case with the above issue).
```
hadley, line 65:
```suggestion
We believe that this usage of `ggsave()` is the off-label use that Hadley talks about in his keynote. Off-label use of functions comprise of using functions in a way that only work *by accident*, and are thus susceptible to breakage at any point due to changes in the code. Another common word for this is "a hack", but this term can often imply that the user is full aware of the brittle nature of the setup. Off-label use can just as well be passed on between users to a point where some thinks that this is the correct, supported, way of doing things (this was certainly the case with the above issue).
```

content/blog/off-label-uses-in-ggplot2/index.Rmd
```diff
@@ -0,0 +1,100 @@
...
+
+```{r, eval=FALSE}
+try(
+  ggplot(mpg) + 
+    geom_point(aes(x = displ, y = hwy)) + 
+    ggsave("my_mpg_plot.png")
+)
+```
+
+Another tidbit we can get is that the perceived feature was, even when it worked, extremely brittle. Consider the following code:
```
hadley, line 83:
```suggestion
Another tidbit is that the perceived feature was extremely brittle, even when it worked. Consider the following code:
```

hadley: This is a great point.

hadley (approved):

content/blog/off-label-uses-in-ggplot2/index.Rmd
thomasp85: Ah - I forgot... thanks

PR #531, opened 2021-09-02
Fix typos and some other small tweaks

jennybc:
Sorry I didn't re-knit but I don't have all the necessary packages installed rn.


mine-cetinkaya-rundel (approved):
Thanks, will knit and merge.

PR #536, opened 2021-09-16
Q3 tidymodels roundup

juliasilge:
So... I confused myself 🤪 thinking I was two weeks late for a Q3 roundup post but actually I am two weeks early. I am setting this up as a draft PR for us now to take a look at but I'm going to wait at least a week, and maybe see if we get recipes in to add to this post.

**Note to self:** check for new releases since today and also re-run the tidy thanks functions before publishing.

I plan to come back to this the beginning of the week of September 27 to publish around then.


juliasilge:
@hfrick No rush on this (see above) but want to take a look at the section on dials?

content/blog/tidymodels-2021-q3/index.Rmd
```diff
@@ -0,0 +1,147 @@
...
+
+One of the reasons we found the OkCupid dataset useful was that it included multiple text columns per observation, so removing these two motivated us to look for a new option to include instead. We landed on [metadata for modern artwork from the Tate Gallery](https://modeldata.tidymodels.org/reference/tate_text.html); if you used `okc_text` in the past, we recommend switching to `tate_text`. For example, we can count how many of these examples of artwork involve paper or canvas.
```
hfrick, line 108:
```suggestion
One of the reasons we found the OkCupid dataset useful was that it included multiple text columns per observation, so removing these two datasets motivated us to look for a new option to include instead. We landed on [metadata for modern artwork from the Tate Gallery](https://modeldata.tidymodels.org/reference/tate_text.html); if you used `okc_text` in the past, we recommend switching to `tate_text`. For example, we can count how many of these examples of artwork involve paper or canvas.
```

hfrick:
Looks great, thank you! :sparkles:

juliasilge:
Once we merge in tidymodels/tidymodels.org#229, we can publish this blog post. Right now the date says September 28 but we can push it out a little of course.

PR #540, opened 2021-09-29
Add blog post for archive 1.1.0

jimhester:


jimhester:
Does anyone think we need any tweaks to this, or should I merge it as-is?

content/blog/archive-1-1-0/index.Rmd
```diff
@@ -0,0 +1,128 @@
+---
+output: hugodown::hugo_document
+
+slug: archive-1-1-0
+title: archive 1 1 0
+date: 2021-08-23
```
jennybc, line 6: These need updating.

content/blog/archive-1-1-0/index.Rmd
```diff
@@ -0,0 +1,128 @@
...
+archive_write_files("nycflights13.zip", fs::dir_ls("nycflights13"))
+```
+
+```{r}
+archive("nycflights13.zip")
+```
+
+## Reading single files from an archive
+
+`archive_read()` is used to read a single file from an archive. This function returns an R connection, which can be passed to many R functions that takes a connection object as input. All base R file system functions use connections, as well as some packages like readr.
```
jennybc, line 59:
```suggestion
`archive_read()` is used to read a single file from an archive. This function returns an R connection, which can be passed to many R functions that take a connection object as input. All base R file system functions use connections, as well as some packages like readr.
```

jimhester:
Thank you both for reviewing!

PR #542, opened 2021-10-31
cli 3.1.0 post

gaborcsardi:


content/blog/cli-3-1-0/index.Rmd
```diff
@@ -0,0 +1,235 @@
...
+
+```{r eval = FALSE}
+writeLines(code_highlight(deparse(cli::hash_emoji)))
+```
+
+```{asciicast code, echo = FALSE}
+writeLines(code_highlight(deparse(cli::hash_emoji)))
+```
+
+### Human readable hash functions
```
jennybc, line 218: I feel like I'm already seeing these ... in session info, I think? That might make a nice example to illustrate the use case for these.

jennybc (approved):
LGTM!

The only conclusion I gradually came to that you might want to incorporate is which features of the release are aimed at, say, a package maintainer vs. a user. I think the palette control is mostly aimed at a user who want to customize their experience of cli-baked UIs. Whereas the rest might affect how a maintainer uses cli.

gaborcsardi:
Thanks! Indeed the palettes are mostly for end users. Although we should probably emphasize that as a developer, it is best not to mix your own colors, but just use the default 8 + bright.

PR #560, opened 2022-02-20
usemodels 0.2.0

topepo:


juliasilge (approved):
Looks good 🙌

juliasilge:
I tried `hugodown::use_tidy_thumbnails()` on this post @topepo but the square file is not truly square. Can you reprocess the images however you did that?

PR #567, opened 2022-03-11
parsnip + extensions post for March 2022

topepo:


juliasilge (approved):
This looks good to me!

Just as a reminder, some things that need to happen before this publishes is:

- the rules release
- adding an image + `hugodown::use_tidy_thumbnails()`

content/blog/parsnip-update-03-2022/index.Rmd
```diff
@@ -0,0 +1,139 @@
...
+  labs(x = NULL, y = "lab result") +
+  theme_bw() + 
+  theme(
+    axis.text.x = element_text(angle = 90), 
+    legend.position = "top"
+  )
+```
+
+
+In this data set, the goal is to estimate how much of the variation in the lab test is due to the different samples (as it should be since they are different) or measurement noise. The latter term could be associated with day-to-day differences, people-to-people differences etc. It might also be irreducible noise. In any case, we'd like to get estimates of these two sources of variation.
```
hfrick, line 111:
```suggestion
With this data set, the goal is to estimate how much of the variation in the lab test is due to the different samples (as it should be since they are different) or measurement noise. The latter term could be associated with day-to-day differences, people-to-people differences etc. It might also be irreducible noise. In any case, we'd like to get estimates of these two sources of variation. 
```

content/blog/parsnip-update-03-2022/index.Rmd
```diff
@@ -0,0 +1,139 @@
...
+* [ ] `hugodown::use_tidy_thumbnails()`
+* [x] Add intro sentence, e.g. the standard tagline for the package
+* [x] `usethis::use_tidy_thanks()`
+-->
+
+We're delighted to announce the release of [parsnip](https://parsnip.tidymodels.org/) 0.2.1. parsnip is a unified modeling interface for tidymodels. 
+
+This release of parsnip precipitated releases of our parsnip extension packages: baguette, discrim, plsmod, poissonreg, and rules. It also allowed us to release an additional package called multilevelmod (see the section below). We've kept CRAN busy in the last week or so. 
+
+You can see a full list of parsnip changes in the [release notes](https://parsnip.tidymodels.org/news/index.html#parsnip-020).
```
hfrick, line 38: liking to version 0.2.0 is on purpose, right? am asking because the blog post announces version 0.2.1

hfrick (approved):
looks good!

juliasilge:
The rules release is now done ✅

juliasilge:
Question: why does the image included here (and [this page](https://parsnip.tidymodels.org/reference/linear_reg.html)) show brulee as _not_ needing an extension package?

hfrick:
re brulee: I noticed that one too, somehow thinking of brulee as a parsnip extension package. But brulee is the package that connects to torch -  and the brulee _engines_ are actually in parsnip: https://github.com/tidymodels/parsnip/blob/main/R/logistic_reg_data.R#L655

juliasilge:
But it does need brulee to be installed right?

``` r
library(parsnip)
linear_reg() %>% set_engine("brulee") %>% fit(mpg ~ ., data = mtcars)
#> Error in `check_installs()`:
#> ! This engine requires some package installs: 'brulee'
```

<sup>Created on 2022-03-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

content/blog/parsnip-update-03-2022/index.Rmd
juliasilge: Since this post encompasses at least two releases, I changed the link to just go to the NEWS file in general.

PR #576, opened 2022-04-20
recipes extensions post for April 2022

EmilHvitfeldt:
I tried to only hit the highlights within each package.


juliasilge (commented):
I would definitely rethink the order and/or what you have for embed, but it will be good to have this out there! 🙌

content/blog/recipes-update-04-2022/index.Rmd
```diff
@@ -0,0 +1,137 @@
...
+library(embed)
+library(themis)
+library(textrecipes)
+library(modeldata)
+set.seed(1234)
+```
+
+## embed
+
+`step_feature_hash()` is now soft deprecated in embed in favor of `step_dummy_hash()` in textrecipes. The embed version uses TensorFlow, which for some use cases is quite a dependency. One thing to keep an eye out for when moving over is that the textrecipes version uses `num_terms` instead of `num_hash` to denote the number of columns to output.
```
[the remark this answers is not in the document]
EmilHvitfeldt: The embed release didn't really add much this time around that would be interesting for the users. I'll pull it to the end.  My original thought was to order the package "less, more, more"

content/blog/recipes-update-04-2022/index.Rmd
```diff
@@ -0,0 +1,137 @@
...
+library(themis)
+library(textrecipes)
+library(embed)
+library(modeldata)
+set.seed(1234)
+```
+
+## themis
+
+A new step `step_smotenc()` was added thanks to [Robert Gregg](https://github.com/RobertGregg). This step applies the SMOTENC algorithm to synthetically generate observations from minority classes. The SMOTENC method can handle a mix of categorical and numerical predictors, which was not possible using the existing SMOTE method which could only operate on numeric predictors.
```
topepo, line 61:
```suggestion
A new step `step_smotenc()` was added thanks to [Robert Gregg](https://github.com/RobertGregg). This step applies the [SMOTENC algorithm](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C7&q=SMOTENC+&btnG=) to synthetically generate observations from minority classes. The SMOTENC method can handle a mix of categorical and numerical predictors, which was not possible using the existing SMOTE method which could only operate on numeric predictors. 
```

content/blog/recipes-update-04-2022/index.Rmd
```diff
@@ -0,0 +1,137 @@
...
+library(embed)
+library(modeldata)
+set.seed(1234)
+```
+
+## themis
+
+A new step `step_smotenc()` was added thanks to [Robert Gregg](https://github.com/RobertGregg). This step applies the SMOTENC algorithm to synthetically generate observations from minority classes. The SMOTENC method can handle a mix of categorical and numerical predictors, which was not possible using the existing SMOTE method which could only operate on numeric predictors. 
+
+The `hpc_data` illustrates this use case neatly. The data set contains characteristics of HPC Unix jobs and how long they took to run (the outcome `class`). The outcome is not that balanced, with some classes having almost 10 times fewer observations than others. One way to deal with an imbalance like this is to over-sample the minority observations to mitigate the imbalance.
```
topepo, line 63:
```suggestion
The `hpc_data` illustrates this use case neatly. The data set contains characteristics of HPC Unix jobs and how long they took to run (the outcome column is `class`). The outcome is not that balanced, with some classes having almost 10 times fewer observations than others. One way to deal with an imbalance like this is to over-sample the minority observations to mitigate the imbalance. 
```

PR #579, opened 2022-05-25
dbplyr 2.2.0

hadley:


hadley:
@mgirlich would love your feedback if you have a few minutes between now and June 3.

mgirlich:
Very nice write up! 👍

PR #580, opened 2022-06-01
Point out that usethis can take a ref; update your YAML, folks

jennybc:
It seems worth pointing out that usethis will grab `v2` today, but that it can also be directed to any `ref`.

Also, I've definitely encountered the notion that updating to `v2` can be accomplished by doing find-and-replace `v1` --> `v2`, so I think we should be super clear that it goes beyond that.


hadley (approved):

gaborcsardi:
Thanks!

PR #583, opened 2022-06-08
vetiver blog post announcement

juliasilge:
We are doing an announcement of vetiver on the RStudio blog tomorrow (June 9) and we want to do a small announcement here as well, pointing folks to the blog post on rstudio.com. I'll plan to merge later in the day on Thursday, June 9, after the first blog post goes up.

@isabelizimm want to take a look and make any suggestions?


juliasilge:
@batpigandme Want to see if you have any suggestions for this kind of post?

content/blog/announce-vetiver/index.Rmd
```diff
@@ -0,0 +1,117 @@
...
+```
+
+For both R and Python, the `car_mod` object is a fitted model, with parameters estimated using our training data `mtcars`.
+
+
+## Create a vetiver model
+
+We can create a `vetiver_model()` in R or `VetiverModel()` in Python from the trained model; a vetiver model object collects the information needed to store, version, and deploy a trained model.
+
+```{r}
```
batpigandme, line 88: Maybe not for this blog post, but I'm wondering if it would be helpful to have a language indicator for code chunks in posts where we're using more than one language (e.g. Garrick described an Rmd implementation here: https://rstudio.slack.com/archives/C9NK7JY5B/p1654187053132789).

Once we move to Quarto, tabsets would be the obvious way to go. Basically bringing it up here to see if you think it'd be worthwhile, too.

batpigandme (approved):
This is great, Julia! Enough info that it's not just a redirect to the RStudio blog, but not so much that it's redundant. 🕯

batpigandme (approved):

content/blog/announce-vetiver/index.Rmd
isabelizimm: I think this would be useful going forward, we have been spoiled with quarto in other formats!

isabelizimm (commented):
Looks amazing! I think it covers everything and links to all the necessary places 👍

PR #588, opened 2022-07-08
was it three not five improvements?

maelle:
I was looking for an explanation of multi-paragraph titles because of https://community.rstudio.com/t/one-function-giving-a-undocumented-warning/141708/2 :-)


maelle:
(I see only three items in the list, but maybe that's now how I should count)

PR #590, opened 2022-07-15
Fix author links

jimhester:
They were previously linking to the @ usernames, which was resulting in
a 404 if you tried to follow the link.


jimhester:
cc @MichaelChirico

MichaelChirico:
is that a hugo / blogdown issue more generally? or is `@` being interpreted within the tidyverse site?

MichaelChirico (approved):

PR #609, opened 2022-11-23
dplyr 1.1.0 pre-release post

DavisVaughan:
Somewhat polished post. Needs another read through on my end, and:

- [ ] Add a picture
- [ ] Add acknowledgements when we get close to releasing
- [ ] Update the date before merging

### DIFF
--- a/content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
+++ b/content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
@@ -1,0 +1,548 @@
+---
+output: hugodown::hugo_document
+slug: dplyr-1-1-0-is-coming-soon
+title: dplyr 1.1.0 is coming soon
+date: 2022-11-22
+author: Davis Vaughan
+description: >
+    dplyr 1.1.0 is coming soon! This post introduces some of the exciting new
+    features coming in 1.1.0, and includes a call-for-feedback as we finalize
+    the release.
+photo:
+  url: https://unsplash.com/photos/n6vS3xlnsCc
+  author: Kelley Bozarth
+categories: [package] 
+tags: [dplyr]
+editor_options: 
+  chunk_output_type: console
+---
+
+```{=html}
+<!--
+TODO:
+* [ ] Find photo & update yaml metadata
+* [ ] Create `thumbnail-sq.jpg`; height and width should be equal
+* [ ] Create `thumbnail-wd.jpg`; width should be >5x height
+* [ ] `hugodown::use_tidy_thumbnails()`
+* [ ] Add intro sentence, e.g. the standard tagline for the package
+* [ ] `usethis::use_tidy_thanks()`
+-->
+```
+[dplyr](https://dplyr.tidyverse.org/dev/) 1.1.0 is coming soon!
+This is an exciting release for dplyr, incorporating a number of features that have been in flight for years, including:
+
+-   An inline alternative to `group_by()` that implements temporary grouping
+
+-   New join features, such as non-equi joins
+
+-   `arrange()` improvements with character vectors
+
+-   `morph()`, a generalization of `summarise()`
+
+-   Rewrites of "vector" functions using vctrs, including exciting updates to `case_when()`
+
+This pre-release blog post will discuss these new features in more detail.
+We feel pretty confident about them and can't wait for you to try them out, but we welcome any feedback on the [dplyr issues page](https://github.com/tidyverse/dplyr/issues).
+
+You can see a full list of changes in the [release notes](https://dplyr.tidyverse.org/dev/news/index.html).
+There are many additional improvements that couldn't fit in a single blog post!
+
+dplyr 1.1.0 is not on CRAN yet, but you can install the development version from GitHub with:
+
+```{r, eval = FALSE}
+pak::pak("tidyverse/dplyr")
+```
+
+The development version is mostly stable, but is still subject to minor changes before the official release.
+We don't encourage relying on it for production usage, but we would love for you to try out these new features.
+
+```{r setup, warning=FALSE, message=FALSE}
+library(dplyr)
+library(clock)
+set.seed(12345)
+```
+
+## Temporary grouping with `.by`
+
+Verbs that work "by group," such as `mutate()`, `summarise()`, `filter()`, and `slice()`, have gained a powerful new argument, `.by`, which allows for inline and temporary grouping.
+Grouping radically affects the computation of the dplyr verb you use it with, and one of the goals of `.by` is to allow you to place that grouping specification alongside the code that actually uses it.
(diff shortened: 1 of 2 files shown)


jennybc:
Have you considered splitting this up over multiple posts, similar to the lead up to dplyr 1.0.0? This feels like a new things to discuss in one post! It could be multiple posts, dropped at the same time?

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
```diff
@@ -0,0 +1,548 @@
...
+```
+
+These two particular results look the same, but the behavior of `.by` diverges from `group_by()` when multiple group columns are involved:
+
+-   `summarise(.by = )` always returns an ungrouped data frame
+
+-   `group_by() %>% summarise()` peels off 1 layer of grouping by default, and messages you that it is doing so unless you set `.groups`
+
+```{r}
+# Typically you'd follow this up with `ungroup()` or you'd set `.groups = "drop"`
```
hadley, line 106: Probably don't need quite so much code here

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
```diff
@@ -0,0 +1,548 @@
...
+```{r}
+expenses %>%
+  summarise(cost = mean(cost), .by = c(id, region))
+```
+
+You might also recognize that these results aren't returned in exactly the same order.
+`group_by()` always sorts the grouping keys in ascending order, but `.by` retains the original ordering found in the data.
+If you need ordered summaries with `.by`, we recommend calling `arrange()` explicitly before or after summarizing.
+
+While we've focused on using `.by` with `summarise()`, it also works with other verbs, like `mutate()` and `slice_max()`:
```
hadley, line 121:
```suggestion
While here we've focused on using `.by` with `summarise()`, it also works with other verbs, like `mutate()` and `slice_max()`:
```

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
```diff
@@ -0,0 +1,548 @@
...
+
+```{r setup, warning=FALSE, message=FALSE}
+library(dplyr)
+library(clock)
+set.seed(12345)
+```
+
+## Temporary grouping with `.by`
+
+Verbs that work "by group," such as `mutate()`, `summarise()`, `filter()`, and `slice()`, have gained a powerful new argument, `.by`, which allows for inline and temporary grouping.
```
hadley, line 67:
```suggestion
Verbs that work "by group," such as `mutate()`, `summarise()`, `filter()`, and `slice()`, have gained a experimental new argument, `.by`, which allows for inline and temporary grouping.
```

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
```diff
@@ -0,0 +1,548 @@
...
+Grouping radically affects the computation of the dplyr verb you use it with, and one of the goals of `.by` is to allow you to place that grouping specification alongside the code that actually uses it.
+As an added benefit, with `.by` you no longer need to remember to `ungroup()` after `summarise()`, and `summarise()` won't ever message you about how it's handling the groups!
+
+This exciting feature was inspired by [data.table](https://cran.r-project.org/package=data.table), which bakes this idea directly into `[`.
```
hadley, line 71:
```suggestion
This feature was inspired by [data.table](https://cran.r-project.org/package=data.table), which has always used per-operation grouping.
```

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
```diff
@@ -0,0 +1,548 @@
...
+)
+
+employees
+```
+
+One way to start approaching this problem is to look for the party that happened directly before each birthday.
+You can do this with an inequality join:
+
+```{r}
+left_join(employees, parties, join_by(birthday >= party))
```
hadley, line 200: Use the pipe here?

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
```diff
@@ -0,0 +1,548 @@
...
+  region = c("A", "A", "A", "B", "B", "A", "A"),
+  cost = c(25, 20, 19, 12, 9, 6, 6)
+)
+expenses
+```
+
+If I were to ask you to compute the average `cost` per `region`, you'd probably write something like:
+
+```{r}
+expenses %>%
```
hadley, line 87: Base pipe?

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
```diff
@@ -0,0 +1,548 @@
...
+df %>%
+  morph(x = intersect(x, table), .by = g)
+```
+
+One big difference between `summarise()` and `morph()` is that `morph()` always returns an ungrouped data frame, even if the input was a grouped data frame with multiple groups.
+This simplifies `morph()` immensely, as it doesn't need to inherit the `.groups` argument of `summarise()`, and never emits any messages.
+
+We expect that you'll continue to use `summarise()` much more often than `morph()`, but if you ever find yourself applying a function to each group that returns an arbitrary number of rows, `morph()` should be your go-to tool!
+
+## Vector function rewrites
```
hadley, line 428: I'd be tempted to omit this since I'm not sure it needs comment. (You can still use it in the release blog post)

hadley (commented):
Looks great!

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
DavisVaughan: Are you talking about the little paragraph that follows this, or this whole section? Did you not want to talk about `case_when()` in the pre-release post? I intended for this to be an H2 header to set up the three H3 headers about `case_when` `case_match` and `consecutive_id`

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
DavisVaughan: I think the example with multiple groups is important, so I've kept that, but I removed the bullets and generally simplified the language because I think the output tells most of the story

romainfrancois (approved):
Looks great. I agree that the last paragraph is less critical about receiving feedback, and so could be shelved for a (one of the) release post(s).

content/blog/dplyr-1-1-0-is-coming-soon/index.Rmd
hadley: I was thinking the whole section.

DavisVaughan: I removed it

DavisVaughan:
@jennybc since we've chopped this blog post down a bit to the 4 sections that we are "requesting feedback" on, we are going to send this out as 1 post, with some possible follow up posts that might talk about these or other features in even more detail closer to release

PR #619, opened 2023-01-31
dplyr 1.1.0: Per-operation grouping

DavisVaughan:
Post 2 of 4


content/blog/dplyr-1-1-0-per-operation-grouping/index.Rmd
```diff
@@ -0,0 +1,191 @@
...
+author: Davis Vaughan
+description: >
+    dplyr now supports an experimental per-operation grouping syntax. This serves as an
+    alternative to `group_by()` and always returns an ungrouped data frame, meaning that you
+    never need to remember to `ungroup()`.
+photo:
+  url: https://www.pexels.com/photo/fruit-stand-375897/
+  author: Clem Onojeghuo
+categories: [package] 
+tags: [dplyr]
```
hadley, line 15: I just remembered we used a dplyr-1-0-0 tag (https://www.tidyverse.org/tags/dplyr-1-0-0/) for 1.0.0; we should do the same here, and you can add a link to each blog post to see all the posts.

content/blog/dplyr-1-1-0-per-operation-grouping/index.Rmd
```diff
@@ -0,0 +1,191 @@
...
+editor_options: 
+  chunk_output_type: console
+---
+
+Today we are going to look at one of the major new features in [dplyr 1.1.0](https://dplyr.tidyverse.org/news/index.html#dplyr-110), per-operation grouping with [`.by`/`by`](https://dplyr.tidyverse.org/reference/dplyr_by.html).
+Per-operation grouping is an experimental alternative to `group_by()` which is only active within a single dplyr verb.
```
hadley, line 21: Inspired by data.table.

content/blog/dplyr-1-1-0-per-operation-grouping/index.Rmd
```diff
@@ -0,0 +1,191 @@
...
+# But `.by` used whatever order was already there
+transactions2 |>
+  summarise(total = sum(revenue), .by = c(company, year))
+```
+
+Notice that `.by` doesn't re-sort the grouping keys.
+Instead, the previous call to `arrange()` is "respected" in the summary (this is also useful in combination with the new `.locale` argument to `arrange()`).
+
+We expect that most code won't depend on the ordering of these group keys, but it is worth keeping in mind if you are switching to `.by`.
+If you did rely on sorted group keys, you'll need to explicitly call `arrange()` either before or after the call to `summarise(.by =)`.
```
hadley, line 164: We may add a new argument to control this behaviour in the future?

content/blog/dplyr-1-1-0-per-operation-grouping/index.Rmd
```diff
@@ -0,0 +1,191 @@
...
+transactions |>
+  summarise(total = sum(revenue), .by = c(company, year))
+```
+
+There are a few things about `.by` worth noting:
+
+-   The result is always ungrouped, regardless of the number of grouping columns.
+    With `.by`, you never need to remember to call `ungroup()`.
+
+-   We used tidy-select to group by multiple columns.
```
hadley, line 87:
```suggestion
-   We used [tidyselect](https://tidyselect.r-lib.org/reference/language.html) to group by multiple columns.
```

hadley (approved):

PR #620, opened 2023-02-02
Use ! instead of - for negation of tidy selection

yutannihilation:
Congratulations on the long-awaited release of dplyr! Sorry for nitpicking on this great article, but I believe `!` is recommended.


DavisVaughan:
Thanks!

PR #622, opened 2023-02-06
dplyr 1.1.0: `pick()`, `reframe()`, and `arrange()`

DavisVaughan:


DavisVaughan:
I've added an acknowledgements section since this is the last post

I was thinking about doing an overarching post after this one that included the acknowledgements, but I no longer think it is needed. Instead, I think we can link to the 1.1.0 tag in the dplyr pkgdown site like @batpigandme did here https://github.com/tidyverse/dplyr/pull/6684

hadley (approved):

PR #627, opened 2023-03-08
Pipe comparison post

### LINKED ISSUE #626: Blog post about the base pipe
Using this text removed from R4DS:

---

While `|>` and `%>%` behave identically for simple cases, there are a few crucial differences.
These are most likely to affect you if you're a long-term user of `%>%` who has taken advantage of some of the more advanced features.
But they're still good to know about even if you've never used `%>%` because you're likely to encounter some of them when reading wild-caught code.

-   By default, the pipe passes the object on its left-hand side to the first argument of the function on the right-hand side.
    `%>%` allows you to change the placement with a `.` placeholder.
    For example, `x %>% f(1)` is equivalent to `f(x, 1)` but `x %>% f(1, .)` is equivalent to `f(1, x)`.
    R 4.2.0 added a `_` placeholder to the base pipe, with one additional restriction: the argument has to be named.
    For example, `x |> f(1, y = _)` is equivalent to `f(1, y = x)`.

-   The `|>` placeholder is deliberately simple and can't replicate many features of the `%>%` placeholder: you can't pass it to multiple arguments, and it doesn't have any special behavior when the placeholder is used inside another function.
    For example, `df %>% split(.$var)` is equivalent to `split(df, df$var)` and `df %>% {split(.$x, .$y)}` is equivalent to `split(df$x, df$y)`.

    With `%>%`, you can use `.` on the left-hand side of operators like `$`, `[[`, `[` (which you'll learn about in @sec-subset-many), so you can extract a single column from a data frame with (e.g.) `mtcars %>% .$cyl`.
    A future version of R may add similar support for `|>` and `_`.
    For the special case of extracting a column out of a data frame, you can also use `dplyr::pull()`:

    ```{r}
    mtcars |> pull(cyl)
    ```

-   `%>%` allows you to drop the parentheses when calling a function with no other arguments; `|>` always requires the parentheses.

-   `%>%` allows you to start a pipe with `.` to create a function rather than immediately executing the pipe; this is not supported by the base pipe.

Luckily there's no need to commit entirely to one pipe or the other --- you can use the base pipe for the majority of cases where it's sufficient and use the magrittr pipe when you really need its special features.

batpigandme:
* Closes #626 

@hadley, I left you as author, since 90% of it is your exact language from R4DS. Obviously feel free to change at will.


content/blog/base-vs-magrittr-pipe/index.Rmd
```diff
@@ -0,0 +1,68 @@
...
+* [x] `hugodown::use_tidy_thumbnails()`
+* [ ] Add intro sentence, e.g. the standard tagline for the package
+* [ ] `usethis::use_tidy_thanks()`
+-->
+
+**Note:** The following has been adapted from a section of the forthcoming second edition of [R for Data Science](https://r4ds.hadley.nz/) that had to be removed due to length limitations.
+
+## Pipes
+
+With the 4.1.0 release of R, a native pipe operator, `|>`, was introduced.
```
tracykteal, line 35: Is there a canonical 'we introduced `|>` in R' post that we can link to?

tracykteal: Could we add in a sentence or two here about why a native pipe operator was introduced? I think having a little of that motivation will help people understand how they might be thinking about what to use when, and why differences are important.

content/blog/base-vs-magrittr-pipe/index.Rmd
```diff
@@ -0,0 +1,68 @@
...
+
+    ```{r eval=FALSE}
+    mtcars |> pull(cyl)
+    ```
+
+-   `%>%` allows you to drop the parentheses when calling a function with no other arguments; `|>` always requires the parentheses.
+
+-   `%>%` allows you to start a pipe with `.` to create a function rather than immediately executing the pipe; this is not supported by the base pipe.
+
+Luckily there's no need to commit entirely to one pipe or the other --- you can use the base pipe for the majority of cases where it's sufficient and use the magrittr pipe when you really need its special features.
```
tracykteal, line 68: I feel like this is going to be the main take home for most people - that you can use base pipe most of the time, and magrittr in special cases. 
This is a short enough post that this sentence should stand out being at the end, but another option could be to move this to just before the ## `|>` vs. `%>%` header.

tracykteal:
Looks good! I just had a couple notes about adding a bit more context.

EmilHvitfeldt:
Would this be the place to talk about using native pipes in packages? Using `|>` in an example or in the code itself makes it so the package requires `R (>= 4.1.0)` which affects its reverse dependencies.

This can be a pain as it sometimes feels like a package is bumped from 3.6.0 to 4.1.0, seemingly only for a couple of `|>`s.

hadley:
Yeah, I was thinking we might talk a little bit about what we did in purrr 1.0.0 to make the pipe work in the docs, and how you can expect to see this sort of pattern in more tidyverse packages in the future.

content/blog/base-vs-magrittr-pipe/index.Rmd
batpigandme: There are a bunch of posts on the introduction of the native pipe, but (and I've looked), I can't find one that's "official," per se. Jumping Rivers' post, [New features in R 4.1.0](https://www.jumpingrivers.com/blog/new-features-r410-pipe-anonymous-functions/), Michael Browman gets points for posting about it when it was still in R-devel ([The New Base Pipe](https://michaelbarrowman.co.uk/post/the-new-base-pipe/)), very handy comparison table in an SO thread: [[What are the differences between R's new native pipe `|>` and the magrittr pipe `%>%`?](https://stackoverflow.com/questions/67633022/what-are-the-differences-between-rs-new-native-pipe-and-the-magrittr-pipe)](https://stackoverflow.com/a/72086492/785190).

Because there wasn't a canonical post from R Core, I wasn't sure where to link to. It's in the [R devel news](https://cran.r-project.org/doc/manuals/r-devel/NEWS.html), but there's no way to link directly to the new features section of the 4.1.0 release…

hadley: Yeah, I think quoting from the news is the best way to go.

batpigandme:
Apologies for the messed-up rebase. LMK if you want me to just submit this as a fresh PR.

@hadley, could you take a quick look at the section re. how we used native pipes in purrr? Want to make sure I'm not misrepresenting or saying anything controversial.

content/blog/base-vs-magrittr-pipe/index.Rmd
```diff
@@ -0,0 +1,92 @@
...
+    ```
+
+-   `%>%` allows you to drop the parentheses when calling a function with no other arguments; `|>` always requires the parentheses.
+
+-   `%>%` allows you to start a pipe with `.` to create a function rather than immediately executing the pipe; this is not supported by the base pipe.
+
+## Using the native pipe in packages
+
+Because the native pipe wasn't introduced until 4.1.0, code using `|>` in function reference examples or vignettes will not work on older versions of R, as it is not valid syntax.
+
```
hadley, line 77: This is a problem for the tidyverse because our versioning policies (insert link) mean that our packages need to work on R 3.5.0 and later.

content/blog/base-vs-magrittr-pipe/index.Rmd
```diff
@@ -0,0 +1,92 @@
...
+Does this mean that you need to increase the minimum R version your package depends on in order to use `|>`?
+Not necessarily.
```
hadley, line 79:
```suggestion
Not necessarily: there are two techniques we can use to keep vignettes and examples working.
```

content/blog/base-vs-magrittr-pipe/index.Rmd
```diff
@@ -0,0 +1,92 @@
...
+## Using the native pipe in packages
+
+Because the native pipe wasn't introduced until 4.1.0, code using `|>` in function reference examples or vignettes will not work on older versions of R, as it is not valid syntax.
+
+Does this mean that you need to increase the minimum R version your package depends on in order to use `|>`?
+Not necessarily. 
+
+For example, the base pipe is used in purrr's 1.0.0 release. As can be seen in the [source for the “purrr <-> base R” vignette](https://github.com/tidyverse/purrr/commit/df4630c6e8cd5028386ee96b9036f1755f26adc4), certain code chunks are evaluated conditionally based on the version of R being used. The setup chunk for the vignette includes: `modern_r <- getRversion() >= "4.1.0"`. The results of this are then used in the `eval` argument to determine whether or not a code chunk that relies on "modern R" syntax should be run, as can be seen in the following example:
+
+````{verbatim}
```
hadley, line 83:
```suggestion
````
```

I don't think that's needed.

content/blog/base-vs-magrittr-pipe/index.Rmd
```diff
@@ -0,0 +1,85 @@
...
+
+While `|>` and `%>%` behave identically for simple cases, there are a few crucial differences.
+These are most likely to affect you if you're a long-term user of `%>%` who has taken advantage of some of the more advanced features.
+But they're still good to know about even if you've never used `%>%` because you're likely to encounter some of them when reading wild-caught code.
+
+-   By default, the pipe passes the object on its left-hand side to the first argument of the function on the right-hand side.
+    `%>%` allows you to change the placement with a `.` placeholder.
+    For example, `x %>% f(1)` is equivalent to `f(x, 1)` but `x %>% f(1, .)` is equivalent to `f(1, x)`.
+    R 4.2.0 added a `_` placeholder to the base pipe, with one additional restriction: the argument has to be named.
+    For example, `x |> f(1, y = _)` is equivalent to `f(1, y = x)`.
```
hadley, line 57: R 4.3.0 adds an additional feature that allows you to use `_` in an extraction call like `_$x` or `_[["y"]]`.

content/blog/base-vs-magrittr-pipe/index.Rmd
```diff
@@ -0,0 +1,85 @@
...
+
+-   The `|>` placeholder is deliberately simple and can't replicate many features of the `%>%` placeholder: you can't pass it to multiple arguments, and it doesn't have any special behavior when the placeholder is used inside another function.
+    For example, `df %>% split(.$var)` is equivalent to `split(df, df$var)`, and `df %>% {split(.$x, .$y)}` is equivalent to `split(df$x, df$y)`.
+
+    With `%>%`, you can use `.` on the left-hand side of operators like `$`, `[[`, `[` , so you can extract a single column from a data frame with (e.g.) `mtcars %>% .$cyl`.
+    A future version of R may add similar support for `|>` and `_`.
```
hadley, line 63:
```suggestion
    R added support for this feature in R 4.3.0
```

hadley (approved):

PR #630, opened 2023-03-13
Post for C/C++ issues on CRAN

ateucher:
@gaborcsardi and @DavisVaughan both had a look at an earlier draft.


content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+```
+
+```{r}
+#| eval: false
+say_height(1824) # This will abort due to buffer overflow
+```
+
+### How to fix it
+
+In most cases, this should be a simple fix: replace <code>sprintf()</code> with `snprintf()` and `vsprintf()` with `vsnprintf()`. These `n` variants take a second parameter that specifies the maximum number of bytes written. If the output is a static buffer, you can use `sizeof()`:
```
jennybc, line 126: Maybe also explain the return value of `snprintf()`? Since it does come up below.

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+    }
+    return n;
+  }
+')
+
+say_height_safely(182)
+say_height_safely(1824)
+```
+
+If the destination is not a static buffer, the easiest thing to do is pass in the
```
jennybc, line 148: I had cases that were harder to analyze, but I don't have anything productive to say other than "sometimes it's kind of hard to figure this out" 😬

jennybc (approved):
A few comments, but in general LGTM!

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+    // "My height is xxx cm" is 19 characters but we need 
+    // to add one for the null-terminator
+    char out[19 + 1];
+    int n;
+    n = snprintf(out, sizeof(out), "My height is %i cm", height);
+    Rprintf(out);
+    if (n >= sizeof(out)) {
+       Rprintf("\\nTruncated because input is longer than allowed!\\n");
+    }
+    return n;
```
hadley, line 140: This is probably overkill

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+};
+```
+
+Here is an example where the authors of [Cubist](https://topepo.github.io/Cubist/) applied the [necessary patches](https://github.com/topepo/Cubist/pull/46), and [another one in rlang](https://github.com/r-lib/rlang/pull/1508).
+
+### Vendored code
+
+Function declarations without a prototype are very common, and unfortunately are thus likely to appear in libraries that you include in your package. This may require you to patch that code in your package. The [readxl](https://readxl.tidyverse.org) package includes the [libxls C library](https://github.com/libxls/libxls), which was patched [in readxl here](https://github.com/tidyverse/readxl/commit/afdc9b90cfc2bb1e1c5490c7ba3af5ecfc4a7876) to deal with this issue.
+
+The ideal solution in cases like this would be to submit patches to the upstream libraries so you don't have to deal with the ongoing maintenance of your local patches, but that is not always possible.
```
hadley, line 205: Generally, however, you can explain this problem when submitting your package, and as long as you've have notified the upstream maintainer, CRAN will accept your updated package.

hadley (approved):
Looks great!

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+* [x] Create `thumbnail-sq.jpg`; height and width should be equal
+* [x] Create `thumbnail-wd.jpg`; width should be >5x height
+* [x] `hugodown::use_tidy_thumbnails()`
+* [ ] `Add intro sentence, e.g. the standard tagline for the package
+* [ ] `usethis::use_tidy_thanks()`
+-->
+
+The R package landscape is dynamic, with changes in infrastructure common, especially when CRAN makes changes to their policies and requirements. This is particularly true for packages that include low-level compiled code, requiring developers to be nimble in responding to these changes.
+
+The tidyverse team at Posit is in the unique situation where we have a concentration of skilled developers whose job it is to create and maintain open source packages. This internal community provides the opportunity to collaborate to develop best practices and discover solutions to problems that arise. When we can, we like to share what we've learned so other developers can benefit.
```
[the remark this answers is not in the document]
tracykteal: I'd also suggest 'best practices' -> 'shared practices'

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+
+-   You can use a configure script at the top level of the package, and have it add `CXX_STD=CXX11` for R 3.5 and below. An example (unmerged) [pull request to the readxl](https://github.com/tidyverse/readxl/pull/722/files) package demonstrates this approach. You will also need to add `Biarch: true` in your DESCRIPTION file. This appears to be the approach preferred by CRAN.
+-   For users with R <= 3.5 on a system with an older compiler, package authors can instruct users to edit their `~/.R/Makevars` file to include this line: `CXX_STD=CXX11`.
+
+The tidyverse has a [policy of supporting four previous versions](https://www.tidyverse.org/blog/2019/04/r-version-support/) of R. Currently that includes R 3.5, but with the upcoming release of R 4.3 the minimum version we will support is R 3.6. As we won't be supporting R 3.5 in the near future, you should not feel pressured to either unless you have a compelling reason.
```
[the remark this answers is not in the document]
gaborcsardi: The date of 4.3.0 is not known yet.

tracykteal (approved):
Looks good to me too!

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
+---
+output: hugodown::hugo_document
+
+slug: cran-checks-compiled-code
+title: "New CRAN Requirements for Packages with C and C++"
```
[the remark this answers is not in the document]
ateucher: May as well be consistent. Thanks!

content/blog/cran-checks-compiled-code/index.Rmd
ateucher: Do you have any examples you could point me to that might help me come up with a better example as per @hadley's [comment](https://github.com/tidyverse/tidyverse.org/pull/630#discussion_r1135822319)?

content/blog/cran-checks-compiled-code/index.Rmd
ateucher: Do you mean the "Truncated..." Rprintf output?

content/blog/cran-checks-compiled-code/index.Rmd
jennybc: I don't think you can use this directly, but it's an example of a more complicated situation. Here's a patch in the grisu code that is vendored into readr and vroom.

https://github.com/tidyverse/readr/commit/9143718735136c05c52c467b996dff799f77b888

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+    char out[size]; 
+    int n; 
+    n = snprintf(out, size, "My height is %i cm", height);
+    Rprintf(out);
+    if (n >= sizeof(out)) {
+       Rprintf("\\nTruncated because input is longer than allowed!\\n");
+    }
+    return n;
+  }
+')
```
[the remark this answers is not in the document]
ateucher: I pushed a new example... don't know if it's too contrived?

content/blog/cran-checks-compiled-code/index.Rmd
jennybc: And here's where it intersects with code we control:

https://github.com/tidyverse/readr/blob/9558d899b7e8c592be339bfc31ec35593ccf20ca/src/write_delim.cpp#L219

So I guess I could have hardwired a size of 32, but that would have been really opaque on the grisu side.

jennybc: This is the problem with these mandates, because the grisu function should really be refactored to "know" about the size of buffer that's available to it. But with vendored code, you want to make the smallest patch possible.

ateucher: That's great, thanks! I think it's a great example because you do a nice job explaining the choices you made. I did also just push up a new example, but I think it's pretty contrived. I could link to this instead (or as well)?

content/blog/cran-checks-compiled-code/index.Rmd
hadley: Yeah

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+
+-   You can use a configure script at the top level of the package, and have it add `CXX_STD=CXX11` for R 3.5 and below. An example (unmerged) [pull request to the readxl](https://github.com/tidyverse/readxl/pull/722/files) package demonstrates this approach. You will also need to add `Biarch: true` in your DESCRIPTION file. This appears to be the approach preferred by CRAN.
+-   For users with R <= 3.5 on a system with an older compiler, package authors can instruct users to edit their `~/.R/Makevars` file to include this line: `CXX_STD=CXX11`.
+
+The tidyverse has a [policy of supporting four previous versions](https://www.tidyverse.org/blog/2019/04/r-version-support/) of R. Currently that includes R 3.5, but with the upcoming release of R 4.3 (which should be this Spring some time) the minimum version we will support is R 3.6. As we won't be supporting R 3.5 in the near future, you should not feel pressured to either unless you have a compelling reason.
```
hadley, line 74:
```suggestion
The tidyverse has a [policy of supporting four previous versions](https://www.tidyverse.org/blog/2019/04/r-version-support/) of R. Currently that includes R 3.5, but with the upcoming release of R 4.3 (which should be this Spring some time) the minimum version we will support is R 3.6. As we won't be supporting R 3.5 in the near future, you should not feel pressured to either.
```

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+
+These are considered to be a security risk because they potentially allow [buffer overflows](https://en.wikipedia.org/wiki/Buffer_overflow) that write more bytes than are available in the output buffer. This is a risk if the text that is being passed to <code>sprintf()</code> comes from an uncontrolled source.
+
+Here is a very simple example:
+
+```{r}
+library(cpp11)
+
+cpp_function('
+  int say_height(int height) {
```
hadley, line 103: Even here I think capturing the `n` is a bit unusual. It's probably best practice, but I think few people do it, and you get to the heart of the problem faster without it.

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+say_height_safely(18)
+say_height_safely(1824)
+```
+
+Notice that the return value of `sprintf()` and `snprintf()` are slightly different. `sprintf()` 
+returns the total number of characters written (excluding the null-terminator), while `snprintf()`
+returns the length of the formatted string, whether or not it has been truncated to match `size`.
+
+It is a bit trickier if the destination is not a static buffer, so you must set a maximum `size` that you 
+think is reasonable:
```
hadley, line 152:
```suggestion
It is a bit trickier if the destination is not a static buffer, so you'll have to determine the maximum `size` by carefully thinking about the code:
```

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+In most cases, this should be a simple fix: replace <code>sprintf()</code> with `snprintf()` and `vsprintf()` with `vsnprintf()`. These `n` variants take a second parameter `size`, that specifies the maximum number of bytes to be written, *including the automatically appended null-terminator*. If the output is a static buffer, you can use `sizeof()`:
+
+```{r}
+cpp_function('
+  int say_height_safely(int height) {
+    // "My height is xxx cm" is 19 characters but we need 
+    // to add one for the null-terminator
+    char out[19 + 1];
+    int n;
+    n = snprintf(out, sizeof(out), "My height is %i cm\\n", height);
```
hadley, line 134: That string is length 20, so there's no room for the null terminator?

content/blog/cran-checks-compiled-code/index.Rmd
hadley: The example is better, thanks!

content/blog/cran-checks-compiled-code/index.Rmd
ateucher: Fair enough. I was doing it to demonstrate the return value vs the `size` parameter - I've seen places where people capture it and compare it to the expected `size` and then condition on that...

content/blog/cran-checks-compiled-code/index.Rmd
jennybc: I'm not sure this readr/grisu thing is a great example because it's still basically static and, ideally, **we** would have captured the rationale for 32 bytes in **our** code. But we didn't.

Maybe @hadley is thinking more of an example where the size of the buffer depends on some data that can't be known in advance. This is cobbled together from reading from SO threads, so if you use it, make sure to get someone else to affirm it's not too dorky!

``` r
library(cpp11)
cpp_function('
  size_t say_hello(std::string name) {
    std::string fmt("Hello, %s!\\n");
    // subtract 2 for %s, add 1 for null terminator
    size_t size = fmt.size() - 2 + name.size() + 1;
    std::string buf;
    buf.resize(size);
    snprintf(&buf[0], size, fmt.c_str(), name.c_str());
    Rprintf(buf.c_str());
    return size;
  }
')
say_hello("Jenny")
#> Hello, Jenny!
#> [1] 15
```

<sup>Created on 2023-03-14 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

hadley: @jennybc I don't love that code either, because if you're using C++, I think you'd generally do something other than `sprintf()`. I can see places where we might use code like that, but it feels a bit artificial.

Maybe we can just omit an example for the case of a dynamic buffer since it will depend so much on the specific scenario.

jennybc: I searched r-lib and tidyverse packages for `snprintf()` and we really don't have any organic examples of a true dynamic buffer.

jennybc: @ateucher If we're going to keep the link and reference to the readr/grisu thing, I'd like the chance to slightly rephrase the text.

ateucher: I'm not at all opposed to taking out the dynamic buffer example - I don't think the intent is to show people _how_ to figure out the size of something, just to let them know that they need to with this change. It's also at risk of just getting too long...

@jennybc I'll leave it up to you if we keep it in - and definitely please go ahead and edit it if you think it's worth it. Though I think you have a good point about doing some gymnastics so as to not refactor it properly, so maybe it's not the best illustration?

jennybc: Let's just drop it, because it's such a mixed bag.

content/blog/cran-checks-compiled-code/index.Rmd
ateucher: If you're ok with it @hadley I'd rather keep the `n` - I think it helps to demonstrate the slightly different return values of `sprintf()` vs `snprintf()`.

content/blog/cran-checks-compiled-code/index.Rmd
```diff
@@ -0,0 +1,225 @@
...
+    A few recent changes in CRAN requirements for packages containing C or C++
+    code have caused package developers some headhaches. This post outlines
+    the issues and provides solutions the tidyverse team has used to address them.
+
+photo:
+  url: https://unsplash.com/photos/dhGFLj3rI0Q
+  author: Quino Al
+
+# one of: "deep-dive", "learn", "package", "programming", "roundup", or "other"
+categories: [learn, programming]
```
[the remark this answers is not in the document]
ateucher: Should we try to do this for this post? The icons are here: https://github.com/tidyverse/tidyverse.org/tree/main/themes/hugo-graphite/static/css/images and the categories are defined here: https://github.com/tidyverse/tidyverse.org/blob/72bb2dd7ca0b2f211498a891aa54f55ddcad5014/themes/hugo-graphite/static/css/main-site.css#L835-L856

content/blog/cran-checks-compiled-code/index.Rmd
hadley: I don't have strong feelings about it so go ahead.

ateucher: Thanks!

PR #633, opened 2023-04-13
add post on performant tidy interfaces

simonpcouch:
Deploy preview: https://deploy-preview-633--tidyverse-org.netlify.app/blog/2023/04/performant-packages/


content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
+---
+output: hugodown::hugo_document
+
+slug: performant-packages
+title: Writing Performant Code with Tidy Tools
```
simonpcouch, line 5: Not feeling sure how to title this one--would something like "10 Tips for..." cue more clearly the format of this post? Or something else?

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+) %>% 
+  select(expression, median)
+```
+
+- Building a tibble from scratch using `tibble()` actually takes quite a while as well. `tibble()` handles vector recycling and name checking, builds columns sequentially, all that good stuff. If you need that, use `tibble()`, but if you're building a tibble from well-understood inputs, use `new_tibble(list())`. For a middle ground between `tibble()` and `new_tibble(list())` in terms of both performance and safety, use the `df_list()` function from the vctrs package in place of `list()`.
+
+```{r bm-tibble}
+mark(
+  tibble = tibble(a = 1:2, b = 3:4),
+  new_tibble_df_list = new_tibble(df_list(a = 1:2, b = 3:4), nrow = 2),
```
simonpcouch, line 389: I see that `df_list()` itself is super quick, but that `new_tibble()` `unclass()`ing its input results in most of that slowdown. Is this not a pattern we should be recommending?

simonpcouch (commented):
These results were generated with dev versions of dplyr, tibble, purrr, tidyr, tidyselect, rlang, and vctrs as of today. Do we want to any sort of session info recording or is that overboard for this context?

A few points where I'd especially appreciate some perspective:

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+```
+
+As for column-binding:
+
+```{r bm-bind-cols}
+tbl <- tibble(year = rep(1974L, nrow(mtcars_tbl)))
+
+mark(
+  dplyr = bind_cols(mtcars_tbl, tbl),
+  vctrs = vec_cbind(mtcars_tbl, tbl)
```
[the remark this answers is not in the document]
EmilHvitfeldt: I don't love the `bind_rows()` behavior where it converts lists of lists into a data frame - we wish we could remove it.

_Originally posted by @DavisVaughan in https://github.com/tidymodels/yardstick/pull/405#discussion_r1163269673_

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+
+Some examples of changes like this made to tidymodels packages: [tidymodels/parsnip#945](https://github.com/tidymodels/parsnip/pull/932), [tidymodels/parsnip#934](https://github.com/tidymodels/parsnip/pull/934), [tidymodels/parsnip#929](https://github.com/tidymodels/parsnip/pull/929), [tidymodels/parsnip#923](https://github.com/tidymodels/parsnip/pull/923), [tidymodels/parsnip#902](https://github.com/tidymodels/parsnip/pull/902), [tidymodels/dials#277](https://github.com/tidymodels/dials/pull/277), and [tidymodels/tune#637](https://github.com/tidymodels/tune/pull/637).
+
+### Becoming Join-Critical
+
+Two truths:
+
+- dplyr joins are a remarkably safe and powerful way to synthesize data sources.
+
+- One ought to ask themselves "does this really need to be a join?" when combining data sources in package code.
```
ateucher, line 405: Another places I see joins where another approach would be better is lookup tables. Character matching as described [here](https://adv-r.hadley.nz/subsetting.html?q=lookup#lookup-tables) is a great alternative. This example is a bit contrived but illustrates the point:

``` r
library(tidyverse)
library(bench)

mtcars_tbl <- as_tibble(mtcars, rownames = "make_model")

add_size <- function() {
  lu <- c("4" = "small", "6" = "medium", "8" = "big")
  mtcars_tbl$size <- unname(lu[as.character(mtcars_tbl$cyl)])
  mtcars_tbl
}

lu_df <- tibble(cyl = c(4,6,8), size = c("small", "medium", "big"))

mark(
  join = left_join(mtcars_tbl, lu_df, by = "cyl"),
  lookup = add_size()
) |> 
  select(expression, median)
#> # A tibble: 2 × 2
#>   expression   median
#>   <bch:expr> <bch:tm>
#> 1 join        451.9µs
#> 2 lookup       25.9µs
```

<sup>Created on 2023-04-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+
+Some ways to intuit whether joins are necessary:
+
+- Does the complete information I need live somewhere farther up the call stack or in a package environment?
+
+    - If so, could I pass it as an argument or retrieve it from a package environment?
+
+    - If not, could it? i.e. if the elements of the join are not indices of a loop, or are indices that could easily be subsetted on after the fact, do the join just once higher up in the call stack.
+
+- Am I using the complete outputted join result or just a portion? If I end up only making use of column names, or values in one column, or pairings between two columns, I may be able to instead use `$.tbl_df` or `[.tbl_df`.
```
ateucher, line 415: Or subset before joining?

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+
+Two truths:
+
+- dplyr joins are a remarkably safe and powerful way to synthesize data sources.
+
+- One ought to ask themselves "does this really need to be a join?" when combining data sources in package code.
+
+Some ways to intuit whether joins are necessary:
+
+- Does the complete information I need live somewhere farther up the call stack or in a package environment?
```
ateucher, line 409: Do you have any examples of this? It may be a bit abstract...

ateucher (approved):
This is really great! I like the consistent structure in each section of the problem,  the solution(s), pros and cons, examples.

Since you are targeting package authors with this, I wonder about saying that directly somewhere. Something like: "Most of these suggestions are best suited to package code, as the performance-readability tradeoff is more likely to be worth it there -- however there may also be cases in analytical code, especially in production and/or with very large data sets, where these tips will be helpful."

From my perspective I think the utility to package devs is clear.

Apologies for some of the nitpicky english comments :)

content/blog/performant-code/index.Rmd
hadley: BTW the title should be sentence case

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+
+```{r sum-unlist}
+sum(unlist(numbers))
+```
+
+You may have some other ideas of how to solve this problem! How do we figure out which one is fastest, though? The `mark()` function from bench takes in different proposals to solve the same problem and returns a tibble with information about how long they took (among other things.)
+
+```{r bm-reduce}
+res <-
+  mark(
```
hadley, line 111: IMO this should be `bench::mark()` as the bench package is really designed to be used without attaching.

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+    approach_1 = Reduce(sum, numbers),
+    approach_2 = sum(unlist(numbers))
+  )
+
+res %>% select(expression, median)
+```
+
+The other nice part about `mark()` is that it will check that each approach gives the same output, so that you don't mistakenly compare apples and oranges.
```
hadley, line 119:
```suggestion
The other nice part about `bench::mark()` is that it will check that each approach gives the same output, so that you don't mistakenly compare apples and oranges.
```

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+
+```{r load-vctrs, message = FALSE, warning = FALSE}
+library(vctrs)
+```
+
+If you use dplyr and tidyr like I do, turns out you're also a vctrs user! dplyr and tidyr rely on vctrs to handle all sorts of elementary operations behind the scenes, and the package is a core part of a tidy developer's toolkit. Taken together with some functions from the tibble package, these tools provide a super efficient, albeit bare-bones, alternative interface to common data manipulation tasks like `filter()`ing and `select()`ing.
+
+## Rewriting tidy code
+
+For every change I make to tidymodels packages that just rewrites dplyr and tidyr code in vctrs and tibble, I make probably two or three making simpler optimizations. Reducing duplicated computations, implementing early returns where possible, using vectorized implementations, and other tooling-agnostic changes will likely take you far when optimizing R code. Profiling is your ground truth! When profiling indicates that otherwise well-factored code is slowed by tidy interfaces, though, all is not lost.
```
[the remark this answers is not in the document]
hadley: Maybe "by rewriting dplyr and tidyr code in vctrs and tibble" -> "by rewriting dplyr and tidyr code to instead use vctrs and tibble" ?

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+
+We'll demonstrate different ways to speed up tidy code using a version of the base R data frame `mtcars` converted to a tibble:
+
+```{r mtcars-tbl}
+mtcars_tbl <- as_tibble(mtcars, rownames = "make_model")
+
+mtcars_tbl
+```
+
+### One-for-One Replacements
```
hadley, line 154:
```suggestion
### One-for-one replacements
```

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+
+#### `filter()`
+
+The dplyr code:
+
+```{r filter, eval = FALSE}
+filter(mtcars_tbl, hp > 100)
+```
+
+...can be substituted with:
```
hadley, line 166:
```suggestion
...can be replaced by:
```

?

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+cols <- c("hp", "wt")
+
+mark(
+  dplyr = select(mtcars_tbl, all_of(cols)),
+  `[.tbl_df` = mtcars_tbl[cols]
+) %>% 
+  select(expression, median)
+```
+
+Note that `[.tbl_df` sets `drop = TRUE`, so no need to manually specify that argument if you're used to doing so for `data.frame`s. Be sure that you're subsetting a tibble rather than a data frame, though.
```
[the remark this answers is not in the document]
hadley: `[.data.frame` also doesn't drop when you use it with 1d subsetting, as here.

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+  new_tibble_list = new_tibble(list(a = 1:2, b = 3:4), nrow = 2)
+) %>% 
+  select(expression, median)
+```
+
+Note that `new_tibble()` *will not check the lengths of its inputs.* Carry out simple recycling yourself, and be sure to use the `nrow` argument to get basic length checks.
+
+Some examples of changes like this made to tidymodels packages: [tidymodels/parsnip#945](https://github.com/tidymodels/parsnip/pull/932), [tidymodels/parsnip#934](https://github.com/tidymodels/parsnip/pull/934), [tidymodels/parsnip#929](https://github.com/tidymodels/parsnip/pull/929), [tidymodels/parsnip#923](https://github.com/tidymodels/parsnip/pull/923), [tidymodels/parsnip#902](https://github.com/tidymodels/parsnip/pull/902), [tidymodels/dials#277](https://github.com/tidymodels/dials/pull/277), and [tidymodels/tune#637](https://github.com/tidymodels/tune/pull/637).
+
+### Becoming Join-Critical
```
hadley, line 399:
```suggestion
### Becoming join-critical
```

hadley (approved):

content/blog/performant-code/index.Rmd
simonpcouch: [a6ca771](https://github.com/tidyverse/tidyverse.org/pull/633/commits/a6ca7716d5da34aa3a5aaca3fc445136901e4633), thanks!

content/blog/performant-code/index.Rmd
simonpcouch: Done in [1e34834](https://github.com/tidyverse/tidyverse.org/pull/633/commits/1e3483410aabc8543c4d0824c42fdd6063ab1e9e), thanks!

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+filter(mtcars_tbl, hp > 100)
+```
+
+...can be substituted with:
+
+```{r filter-alt, eval = FALSE}
+vec_slice(mtcars_tbl, mtcars$hp > 100)
+```
+
+Note that the second argument that determines which rows to keep requires you to actually pass the column `mtcars$hp` rather than its reference `hp`. If you feel cozier with square brackets, you can also use `[.tbl_df`:
```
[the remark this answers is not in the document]
simonpcouch: Wonderful, thank you. How does [6cd39c5](https://github.com/tidyverse/tidyverse.org/pull/633/commits/6cd39c5fa0f153a4dbda3b5d99887efd62376130) feel?

I tend to blend `[.tbl_df` and `vec_slice()` depending on whether I'm also subsetting columns, though it probably would be better to be consistent.

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+```{r median-time, include = FALSE}
+median_time <- res$median[[1]]
+```
+
+There are two important lessons to take in from this output:
+
+* The `sum(unlist())` approach was wicked fast compared to `Reduce()`. 
+* Both of these expressions were fast. Even the slower of the two took `r format(median_time)`---to put that in perspective, that expression could complete `r format(round(1 / as.numeric(median_time)))` iterations in a second! Keeping this bigger picture in mind is always important when benchmarking; if code runs fast enough to not be an issue in practical situations, then it need not be optimized in favor of less readable or safe code.
+
+The results of little experiments like this one can be surprising at first. Over time, though, you will develop intuition for the fastest way to solve problems you commonly solve, and will write fast code the first time around!
```
DavisVaughan, line 126: It might be helpful to say something about the "intuition" in this case. For me, it goes something like:

> In this case, using `Reduce()` means calling `sum()` many times, approximately once for each element of the list, and while `sum()` isn't particularly slow, calling an R function many times tends to have non-negligible overhead. Whereas with the `sum(unlist())` approach, there are only 2 R function calls - one for `unlist()` and one for `sum()`, which both immediately drop into C code.

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+### One-for-one replacements
+
+Many of the core functions in dplyr have alternatives in vctrs and tibble that can be quickly transitioned. There are a couple considerations associated with each, though, and some of them make piping a bit more awkward---most of the time, when I switch these out, I remove the pipe `%>%` as well.
+
+#### `filter()`
+
+The dplyr code:
+
+```{r filter, eval = FALSE}
+filter(mtcars_tbl, hp > 100)
```
DavisVaughan, line 159:
```suggestion
mtcars_tbl %>%
  filter(hp > 100)
```

Just because this is most likely what it will look like when you come across it in practice

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+```
+
+...can be replaced by:
+
+```{r filter-alt, eval = FALSE}
+vec_slice(mtcars_tbl, mtcars$hp > 100)
```
DavisVaughan, line 165:
```suggestion
vec_slice(mtcars_tbl, mtcars_tbl$hp > 100)
```

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+```
+
+Note that the second argument that determines which rows to keep requires you to actually pass the column `mtcars$hp` rather than its reference `hp`. If you feel cozier with square brackets, you can also use `[.tbl_df`:
+
+```{r filter-alt-2, eval = FALSE}
+mtcars_tbl[mtcars$hp > 100, ]
```
DavisVaughan, line 171:
```suggestion
mtcars_tbl[mtcars_tbl$hp > 100, ]
```

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+```
+
+The benchmarks for these different approaches are:
+
+```{r bm-filter}
+res <-
+  bench::mark(
+    dplyr = filter(mtcars_tbl, hp > 100),
+    vctrs = vec_slice(mtcars_tbl, mtcars_tbl$hp > 100),
+    `[.tbl_df` = mtcars_tbl[mtcars$hp > 100, ]
```
DavisVaughan, line 183:
```suggestion
    `[.tbl_df` = mtcars_tbl[mtcars_tbl$hp > 100, ]
```

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+filter(mtcars_tbl, hp > 100)
+```
+
+...can be replaced by:
+
+```{r filter-alt, eval = FALSE}
+vec_slice(mtcars_tbl, mtcars$hp > 100)
+```
+
+Note that the second argument that determines which rows to keep requires you to actually pass the column `mtcars$hp` rather than its reference `hp`. If you feel cozier with square brackets, you can also use `[.tbl_df`:
```
DavisVaughan, line 168:
```suggestion
Note that the second argument that determines which rows to keep requires you to actually pass the column `mtcars_tbl$hp` rather than its reference `hp`. If you feel cozier with square brackets, you can also use `[.tbl_df`:
```

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+
+```{r bm-mutate}
+bench::mark(
+  dplyr        = mutate(mtcars_tbl, year = 1974L),
+  `$<-.tbl_df` = {mtcars_tbl$year <- 1974L; mtcars_tbl}
+) %>%
+  select(expression, median)
+```
+
+By default, both `mutate()` and `$<-.tbl_df` append the new column at the right-most position. The `.before` and `.after` arguments to `mutate()` are a really nice interface to adjust that behavior, and I miss it often when using `$<-.tbl_df`. In those cases, `select()` and its alternatives (see next section!) can be helpful.
```
DavisVaughan, line 222: I would also say something like

> Note that we weren't using dplyr's grouped `mutate()` capabilities in this example. In cases where you rely on grouped expressions, the translation from dplyr to vctrs would get much more complex, and is often not worth the switch.

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+bench::mark(
+  dplyr = select(mtcars_tbl, all_of(cols)),
+  `[.tbl_df` = mtcars_tbl[cols]
+) %>% 
+  select(expression, median)
+```
+
+Note that `[.tbl_df` always sets `drop = FALSE`.
+
+`[.tbl_df` can also be used as an alternative interface to `select()` with a `.before` or `.after` argument as well as `relocate()`. For instance, to place that column `year` we made in the last section as the second column, we could write:
```
DavisVaughan, line 264:
```suggestion
`[.tbl_df` can also be used as an alternative interface to `select()` or `relocate()` with a `.before` or `.after` argument. For instance, to place that column `year` we made in the last section as the second column, we could write:
```

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,530 @@
...
+In general, the introduction of groups makes these substitutions much trickier. In those cases, it's likely best to weigh (via profiling) how significant the slowdown is and, if it's not too bad, opt not to make any changes. For code that relies on `group_by()` and sees heavy traffic, see `vctrs::list_unchop()`, `vctrs::vec_chop()`, and `vctrs::vec_rep_each()`.
+
+### Tibbles
+
+Tibbles are great, and I don't want to interface with any other data frame-y thing. Some notes:
+
+- `as_tibble()` on a tibble is not "free":
+
+```{r bm-as-tibble}
+mark(
```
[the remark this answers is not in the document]
simonpcouch: Seems to be just about constant:

``` r
library(tibble)
library(vctrs)
#> 
#> Attaching package: 'vctrs'
#> The following object is masked from 'package:tibble':
#> 
#>     data_frame

rep_mtcars <- function(replicates) {
  vec_slice(mtcars, rep(1:32, replicates))
}

bench::press(
  replicates = 10^(0:3),
  {
    mtcars_df <- rep_mtcars(replicates)
    mtcars_tbl <- as_tibble(mtcars_df, rownames = "make_model")
    bench::mark(
      on_df = as_tibble(mtcars_df),
      on_tbl = as_tibble(mtcars_tbl, rownames = "make_model"),
      check = FALSE
    )
  }
)
#> Running with:
#>   replicates
#> 1          1
#> 2         10
#> 3        100
#> 4       1000
#> # A tibble: 8 × 7
#>   expression replicates      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>      <dbl> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 on_df               1   47.4µs   50.3µs    19177.        0B     38.7
#> 2 on_tbl              1  234.6µs    246µs     3926.      544B     34.0
#> 3 on_df              10   47.1µs     49µs    20187.        0B     40.3
#> 4 on_tbl             10  234.2µs  244.1µs     4054.      544B     36.1
#> 5 on_df             100   47.2µs   49.7µs    19731.        0B     39.7
#> 6 on_tbl            100  231.6µs  243.4µs     4070.      544B     35.9
#> 7 on_df            1000   46.5µs   48.3µs    20386.        0B     43.5
#> 8 on_tbl           1000  234.3µs  241.4µs     4096.      544B     36.6
```

<sup>Created on 2023-04-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+
+Note that `$` will raise a warning if the subsetted column doesn't exist, while `[[` will silently return `NULL`.
+
+With benchmarks:
+
+```{r bm-pull}
+bench::mark(
+  dplyr      = pull(mtcars_tbl, hp),
+  `$.tbl_df` = mtcars_tbl$hp,
+  `[[.tbl_df` = mtcars_tbl[["hp"]]
```
DavisVaughan, line 331: Alignment of `=` is off. In general i dont think there is too much benefit to trying to add extra spaces to keep the RHSs aligned

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+mtcars_tbl$hp
+```
+
+...or:
+
+```{r pull-alt-2, eval = FALSE}
+mtcars_tbl[["hp"]]
+```
+
+Note that `$` will raise a warning if the subsetted column doesn't exist, while `[[` will silently return `NULL`.
```
DavisVaughan, line 323:
```suggestion
Note that, for tibbles, `$` will raise a warning if the subsetted column doesn't exist, while `[[` will silently return `NULL`.
```

content/blog/performant-code/index.Rmd
DavisVaughan: Oh I see you mention grouping later on. Maybe change the title there to just:

> ## Groupings

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+
+- `as_tibble()` on a tibble is not "free":
+
+```{r bm-as-tibble}
+bench::mark(
+  on_tbl_df = as_tibble(mtcars_tbl),
+  on_data.frame = as_tibble(mtcars, rownames = "make_model")
+) %>% 
+  select(expression, median)
+```
```
DavisVaughan, line 380: I believe you can get the block aligned with the above bullet if you add 3 spaces to the beginning of each line in the block, including the triple backtick lines

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+
+```{r supplement-simons-cars}
+supplement_simons_cars <- function() {
+  res <-   
+    vec_slice(
+      mtcars_tbl, 
+      mtcars_tbl$make_model %in% simons_cars$make_model
+    )
+  
+  res$vibey <- simons_cars$vibey
```
DavisVaughan, line 444: I don't think this is exactly right because it relies on the ordering between the two being the same and on `simons_cars` being 100% represented in `mtcars_tbl`

I'd be tempted to refactor this a little into something that uses `vec_match()` and `vec_cbind()`, and that shows how the `inner_join()` is being mimic-ed by dropping unmatched rows.

This doesn't quite fit your story, but maybe you can make a columns like `dream = c(FALSE, TRUE)` to be like "car i actually owned vs car i owned in a dream" or something similar.

``` r
library(tidyverse)
library(vctrs)

mtcars_tbl <- as_tibble(mtcars, rownames = "make_model")

simons_cars <- 
  tibble(
    make_model = c("Honda Civic", "Honda Accord"),
    vibey = c("Yes", "No")
  )

# Pull in the details about my cars
simons_cars %>%
  inner_join(mtcars_tbl, by = join_by(make_model))
#> # A tibble: 1 × 13
#>   make_model  vibey   mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear
#>   <chr>       <chr> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1 Honda Civic Yes    30.4     4  75.7    52  4.93  1.62  18.5     1     1     4
#> # ℹ 1 more variable: carb <dbl>

# If you expect that each of Simon's cars has exactly 0 or 1 match in
# `mtcars_tbl`, then you can do this instead
loc <- vec_match(simons_cars$make_model, mtcars_tbl$make_model)
loc
#> [1] 19 NA

# Keep only the matches
loc_simons <- which(!is.na(loc))
loc_mtcars <- vec_slice(loc, !is.na(loc))

# Drop duplicated join column
mtcars_tbl_join <- mtcars_tbl[setdiff(names(mtcars_tbl), "make_model")]

vec_cbind(
  vec_slice(simons_cars, loc_simons),
  vec_slice(mtcars_tbl_join, loc_mtcars)
)
#> # A tibble: 1 × 13
#>   make_model  vibey   mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear
#>   <chr>       <chr> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1 Honda Civic Yes    30.4     4  75.7    52  4.93  1.62  18.5     1     1     4
#> # ℹ 1 more variable: carb <dbl>
```

<sup>Created on 2023-04-14 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

DavisVaughan: I still think your comment at the end applies (about how if it got more complex then you may have some issues), but I think this is a little closer to "what I'd actually do" when porting code to vctrs

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+
+`nest()`s are subject to similar considerations as joins. When they allow for expressive or principled user interfaces, use them, but manipulate them sparingly in backends. Writing out `nest()` calls _can_ result in substantial speedups, though, and the process is not quite as gnarly as writing out a join. For code that relies on `nest()`s and sees heavy traffic, rewriting with vctrs may be worth the effort.
+
+For example, consider nesting `mtcars_tbl` by `cyl`:
+
+```{r nest}
+nest(mtcars_tbl, .by = cyl)
+```
+
+For some basic nests, like nesting by values in one column, `vec_split()` can do the trick.
```
DavisVaughan, line 476: `vec_split()` takes data frames so you can nest by more than one column too. I think it would actually be worth showing an example of this where you'd do something like:

```r
vec_cbind(res$key, data_frame(data = res$val))
```

content/blog/performant-code/index.Rmd
```diff
@@ -0,0 +1,526 @@
...
+
+bench::mark(
+  glue = glue::glue("My name is {name}."),
+  paste0 = paste0("My name is ", name, "."),
+  check = FALSE
+) %>% 
+  select(expression, median)
+```
+
+My rule of thumb is to use `glue()` for errors, when the function will stop executing anyway. For simple pastes that are intended to be called repeatedly, use `paste0()`. There's a lot of gray area in between those two contexts---intuit (or profile) as you will.
```
DavisVaughan, line 522: You should warn about `paste0("prefix-", character())` not following tidyverse recycling rules, possibly resulting in lots of bugs. A decent / safer middle ground is:

```r
vec_paste0 <- function (...) {
  args <- vec_recycle_common(...)
  exec(paste0, !!!args)
}
```

DavisVaughan: And if you can require a new enough R version, you can use `paste0(recycle0 = TRUE)`

DavisVaughan (approved):

content/blog/performant-code/index.Rmd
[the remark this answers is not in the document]
ateucher: That looks great to me.

Yeah, that makes sense - maybe if the message is "be consistent except when it doesn't make sense to be consistent" it's fine to leave as is!

content/blog/performant-code/index.Rmd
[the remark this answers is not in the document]
ateucher: Interesting, so I guess it's even less of an issue. Nice!

content/blog/performant-code/index.Rmd
ateucher: Sorry if I'm being a bit thick and missing the point... but is the message really just "avoid doing joins repeatedly and find a way to do it once early and subset later"?

simonpcouch: Yes! Will clarify that language.

PR #645, opened 2023-07-17
Q2 2023 tidymodels roundup

hfrick:
todo:
- [ ]  update date before publication


simonpcouch (approved):
This looks great, thanks for making this happen!🦋

content/blog/tidymodels-2023-q2/index.Rmd
```diff
@@ -0,0 +1,168 @@
...
+```
+
+## modeldatatoo
+
+Many of the datasets used in tidymodels examples are available in the modeldata package. The new modeldatatoo package now extends the collection by several bigger datasets. To allow for the bigger size, the package does not contain those datasets directly but rather provides functions to access them, prefixed with `data_`. To access an old friend, the Ames housing data, call `data_ames()`.
+
+```{r}
+library(modeldatatoo)
+
+data_ames()
```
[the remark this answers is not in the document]
hfrick: yep, I've removed it. thanks!

content/blog/tidymodels-2023-q2/index.Rmd
```diff
@@ -0,0 +1,170 @@
...
+First, loading the collection of packages:
+
+```{r}
+#| eval: FALSE
+library(tidymodels)
+```
+
+## modeldatatoo
+
+Many of the datasets used in tidymodels examples are available in the modeldata package. The new modeldatatoo package now extends the collection by several bigger datasets. To allow for the bigger size, the package does not contain those datasets directly but rather provides functions to access them, prefixed with `data_`. To access an old friend, the Ames housing data, call `data_ames()`.
```
EmilHvitfeldt, line 91: Could we add an additional sentence to clarify the usage of this package? Right now it might confuse people why we added another package.

something the following:

modeldata data sets can be used everywhere, and will most often be used in package documentation. modeldatatoo required an active internet connection to be used and will thus be used mostly for workshops and websites.

content/blog/tidymodels-2023-q2/index.Rmd
```diff
@@ -0,0 +1,170 @@
...
+get_rsplit(ames_folds, 3) |> 
+  autoplot()
+```
+
+The `ames_fold` object can then be used with functions from the tune package as usual.
+
+
+## tidymodels.org
+
+The tidymodels website, [tidymodels.org](https://www.tidymodels.org/), has been updated to use [Quarto](https://quarto.org/). Things largely look the same as before but this change has allowed us to improve the search functionality of the website.
```
EmilHvitfeldt, line 154: And it now uses a build system that allows more people to contribute! 😄

EmilHvitfeldt (approved):
Looking great! I forgot we did some of these things 😄

content/blog/tidymodels-2023-q2/index.Rmd
hfrick: Sure! Anything else you think would be good to add, please feel free! I think I'm lacking a little bit of context for some of these, so very happy for you to add stuff.

I was also wondering why we have the Ames housing data in both packages?

topepo: That's a good question. Actually, it will be in three packages (`AmesHousing` too). Can we remove it here?

content/blog/tidymodels-2023-q2/index.Rmd
```diff
@@ -0,0 +1,173 @@
...
+
+```{r}
+ames_folds
+```
+
+Starting with all other 99 points in the analysis set, points are excluded until the distribution of nearest neighbor distances from the analysis set to the assessment set matches that of nearest neighbor distances from the training set to the prediction sites.
+
+Looking at one of the splits, we can see the single assessment point, the points included in the analysis set, and the points excluded as the buffer.
+
+```{r}
```
topepo, line 147: Totally uptight of me but... can you make the physical size of each axis the same (square figure region)?

content/blog/tidymodels-2023-q2/index.Rmd
EmilHvitfeldt: Sure thing, but it won't hit CRAN for a little bit, because of the pause

EmilHvitfeldt: https://github.com/tidymodels/modeldatatoo/pull/21

content/blog/tidymodels-2023-q2/index.Rmd
hfrick: Hm, I've played around with this but I'd like to show a plain `autoplot()` without tweaks here.

In more detail: These seem like sensible defaults to me, I don't quite see why the data would ask for a square layout, a square layout is a little hard to land on with this coordinate system, and I don't want to divert attention from the content of the plot and the out-of-the-box syntax with the customization.

PR #655, opened 2023-09-03
Update r4ds url

zekiakyol:
- This pull request updates r4ds(<https://r4ds.had.co.nz/>) to r4ds 2e (https://r4ds.hadley.nz/) on the homepage.
- The current amazon link in **:scroll: content\home\band_two.md** also points to the 1st edition (http://amzn.to/2aHLAQ1) but I did not edited that part to not create an accidental referenced link etc.


hadley:
Thanks!

PR #662, opened 2023-11-10
Add scales blogpost

thomasp85:
Awaiting release in 3 weeks


content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+
+# one of: "deep-dive", "learn", "package", "programming", "roundup", or "other"
+categories: [package] 
+tags: []
+---
+
+```{=html}
+<!--
+TODO:
+* [ ] Look over / edit the post's title in the yaml
```
hadley, line 24: Needs updating

content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+
+You can install it from CRAN with:
+
+```{r, eval = FALSE}
+install.packages("scales")
+```
+
+This blog post will give a quick overview of the 1.3.0 release, which is mainly an upkeep release but does contain a few interesting tidbits.
+
+You can see a full list of changes in the [release notes](%7B%20github_release%20%7D)
```
hadley, line 45: Needs update

content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+
+```{r setup}
+library(scales)
+set.seed(1)
+```
+
+## Proper support for difftime objects
+
+While scales have had rudimentary support for objects from the hms package it has not included basic support for difftime objects from base. This is now rectified with the introduction of `label_timespan()`, `breaks_timespan()`, and `transform_timespan()`. While the labels and breaks function can be used on their own, all the behavior is encapsulated in the timespan transform object which is kin to `transform_hms()`.
```
hadley, line 54:
```suggestion
While scales had rudimentary support for objects from the hms package, I did not support the more common base R difftime objects. This is now rectified with the introduction of `label_timespan()`, `breaks_timespan()`, and `transform_timespan()`. While the labels and breaks function can be used on their own, all the behavior is encapsulated in the timespan transform object which is kin to `transform_hms()`.
```

content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+library(ggplot2)
+
+events <- data.frame(
+  time = runif(30, max = 200),
+  magnitude = rnorm(30) + 2
+)
+
+ggplot(events) + 
+  geom_point(aes(time, y = 0, size = magnitude), position = position_jitter(height = 0)) + 
+  scale_x_continuous(trans = transform_timespan(), limits = c(0, NA))
```
hadley, line 66: Will this ever get chosen automatically?

content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+
+## Proper support for difftime objects
+
+While scales have had rudimentary support for objects from the hms package it has not included basic support for difftime objects from base. This is now rectified with the introduction of `label_timespan()`, `breaks_timespan()`, and `transform_timespan()`. While the labels and breaks function can be used on their own, all the behavior is encapsulated in the timespan transform object which is kin to `transform_hms()`.
+
+```{r}
+library(ggplot2)
+
+events <- data.frame(
+  time = runif(30, max = 200),
```
hadley, line 60: Maybe use `as.difftime()` here to more clearly tie it to the new transformers.

content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+)
+
+ggplot(events) + 
+  geom_point(aes(time, y = 0, size = magnitude), position = position_jitter(height = 0)) + 
+  scale_x_continuous(trans = transform_timespan(), limits = c(0, NA))
+```
+
+As we can see the timespan transform by default understand numeric data as seconds elapsed. Further it identifies that for this range, adding breaks for minutes makes most sense.
+
+If we had recorded time as hours rather than seconds, we can inform the transformer of this:
```
hadley, line 71: I think it would be better to illustrate this by setting the `units` attribute.

content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+
+```{r}
+ggplot(events) + 
+  geom_point(aes(time, y = 0, size = magnitude), position = position_jitter(height = 0)) + 
+  scale_x_continuous(trans = transform_timespan("hours"), limits = c(0, NA))
+```
+
+## API brush-up
+
+scales has gone through a number of touch-ups on its API, latest with a revamp of the labels functions which are now all prefixed with `label_`. With this release we continues (and hopefully concludes) this by changing the naming of the transformation utilities and the palettes. More in line with how we cater to tab-completion when designing functions they all now shares a `transform_` and `pal_` prefix respectively. Further, we have moved from the word `trans` to the more descriptive `transform` throughout the API. On top of that we have renamed the `label_dollar()` function to `label_currency()` to make it clear that this can be used for any type of currency, not just US dollars. All the old functions have been kept around with no plan of deprecation but we advice you to update your code to use the new names.
```
hadley, line 81:
```suggestion
scales has gone through a number of touch-ups on its API, such as revamping the labels functions to all start with `label_`. This release we continue (and hopefully conclude) the touch-ups by using a common prefix for the transformation utilities (`transform_`) and palettes (`pal_`). We have also rename `label_dollar()` to `label_currency()` to make it clear that this can be used for any type of currency, not just dollars (US or otherwise). All the old functions have been kept around with no plan of deprecation but we advise you to update your code to use the new names.
```

hadley: Maybe say they've been superseded since that's the official term.

content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+
+## More transformation power
+
+This release also includes some news about transformations apart from the name changes. They have received a fair amount of bug fixes and a new build in transformation type has joined the group. `transform_asinh()`, the inverse hyperbolic sine transformation, can be used much like log transformations, but support negative values as well.
```
hadley, line 85:
```suggestion
This release also includes some other updates to the transformations. They have received a fair amount of bug fixes and a new built-in transformation type has joined the group: `transform_asinh()`, the inverse hyperbolic sine transformation, can be used much like log transformations, but it also supports negative values.
```

content/blog/scales-1-3-0/index.Rmd
```diff
@@ -0,0 +1,101 @@
...
+
+```{r, warning=FALSE}
+plot(transform_asinh(), xlim = c(-100, 100))
+lines(seq(-100, 100), transform_log()$transform(seq(-100, 100)), col = "red")
+text(50, 3, label = "log-transform", col = "red", adj = 0)
+```
+
+Transformation objects can also now optionally contain information about it's derivatives and inverse derivative which makes it possible to properly correct density estimations of transformed values.
```
hadley, line 93:
```suggestion
Transformation objects can now also (optionally) record the derivatives and inverse derivative which makes it possible to properly correct density estimations of transformed values.
```

hadley (approved):

content/blog/scales-1-3-0/index.Rmd
thomasp85: As in, I need to tick off the list?

content/blog/scales-1-3-0/index.Rmd
thomasp85: That would be for ggplot2 to decide but not as it is right now

content/blog/scales-1-3-0/index.Rmd
thomasp85: The point was mostly to show that it is adaptable to how a numeric should be interpreted. But maybe it is better to focus on the difftime object directly

content/blog/scales-1-3-0/index.Rmd
hadley: Yup. That way we all know its been done.

PR #666, opened 2023-11-13
Draft blog post for httr2 1.0.0

hadley:


content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+* [x] Find photo & update yaml metadata
+* [x] Create `thumbnail-sq.jpg`; height and width should be equal
+* [x] Create `thumbnail-wd.jpg`; width should be >5x height
+* [x] `hugodown::use_tidy_thumbnails()`
+* [x] Add intro sentence, e.g. the standard tagline for the package
+* [ ] `usethis::use_tidy_thanks()`
+-->
+```
+We're delighted to announce the release of [httr2](https://httr2.r-lib.org)[^1] 1.0.0.
+httr2 is the second generation of httr; it generates HTTP requests and helps you process the response, design with an towards modern web APIs and potentially putting your code in package.
```
jennybc, line 38: I think the last half of this sentence isn't fully baked yet.

content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+#|   The new httr2 logo is a dark blue hexagon with httr2 written in 
+#|   bright white at the top of logo. Underneath the text is a vibrant
+#|   magenta baseball player hitting a ball emblazoned with the letters
+#|   "www".
+knitr::include_graphics("httr2.png")
+```
+
+httr2 is the successor to httr.
+The big difference is that it has an explicit request object which you can build up over time.
+This makes the interface work much more naturally with pipe, and is generally easier to work with as you can build up a complex request with many simple features.
```
jennybc, line 65: "the pipe" or "pipes"?

content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+If you're a current httr user, there's no need to switch, as we'll continue to maintain the package many years, but if you start on a new project, I'd recommend that you give httr2 a shot.
```
jennybc, line 66:
```suggestion
If you're a current httr user, there's no need to switch, as we'll continue to maintain the package for many years, but if you start on a new project, I'd recommend that you give httr2 a shot.
```

content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+## Error handling
+
+You can use `resp_status()` to see the returned status:
+
+```{r}
+resp |> resp_status()
+```
+
+However, this will almost always be 200, because httr2 automatically follows redirects (values in the 300s) and turns HTTP errors (values in the 400s and 500s) into errors.
+The following example shows what this looks like using a example endpoint that throws the returns the request status:
```
jennybc, line 162: "that throws the returns the request status" still needs some love

content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+
+## Multiple requests
+
+httr2 includes three functions to perform multiple requests:
+
+-   `req_perform_sequential()` takes a list of requests and performs them in order.
+
+-   `req_perform_parallel()` takes a list of requests and performs them in parallel (up to 6 at a time by default).
+    It's similar to `req_perform_sequential()`, but is obviously faster, at the expense of potentially hammering a server.
+    However, it also has some limitations: most importantly it can't re-request an expired OAuth token and it doesn't respect `req_retry()` or `req_throttle()`.
```
jennybc, line 207:
```suggestion
    However, it also has some limitations: most importantly it can't refresh an expired OAuth token and it doesn't respect `req_retry()` or `req_throttle()`.
```

Not sure about this. Are you trying to distinguish the case of "valid refresh token, need new access token" from "invalid refresh token, need to do full dance"?

content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+
+-   `req_perform_iterative()` takes a single request and a callback function to generate the next request from previous response.
+    It'll keep going until it either your callback function returns `NULL` or the `max_reqs` requests have been performed.
```
jennybc, line 210:
```suggestion
    It'll keep going until the callback function returns `NULL` or `max_reqs` requests have been performed.
```

content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+
+```{r}
+resps |> 
+  resps_data(\(resp) list(resp_body_json(resp))) |> 
+  _[1:3] |> 
+  str(list.len = 10)
+```
+
+Another option would be to convert each response into a data frame or tibble.
+That's a little tricky here because of the lists that will need to become list-columns[^2], so we'll avoid that challenge here by focussing the first nine columns:
```
jennybc, line 246:
```suggestion
That's a little tricky here because of the lists that will need to become list-columns[^2], so we'll avoid that challenge here by focussing on the first nine columns:
```

content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+
+```{r}
+sw_data <- function(resp) {
+  tibble::as_tibble(resp_body_json(resp)[1:9])
+}
+resps |> resps_data(sw_data)
+```
+
+When you're performing large numbers of requests, it's almost inevitable that something will go wrong.
+By default, all three functions will bubble up errors, causing you to lose of the work that's been done so far.
```
jennybc, line 259:
```suggestion
By default, all three functions will bubble up errors, causing you to lose all of the work that's been done so far.
```

content/blog/httr2-1-0-0/index.Rmd
```diff
@@ -0,0 +1,268 @@
...
+You can, however, use the `on_error` argument to change what happens, either ignoring errors, or returning when you hit the first error.
+This changes the return value: instead of a list of just responses, the list might also continue error objects.
```
jennybc, line 261:
```suggestion
This changes the return value: instead of a list of just responses, the list might also contain error objects.
```

hadley:
Thanks @jennybc!

PR #667, opened 2023-11-27
fix typo in title

maelle:
@thomasp85 thanks for this great post! (I had not seen the typo either when reading it for the first time last year)

The typo is also in the slug.


thomasp85:
Will this break all the old links?

maelle:
Presumably no because I haven't changed the slug and that's used for creating the URL https://github.com/tidyverse/tidyverse.org/blob/b025086f34a5d38c546ee31ac80783655cc4f46a/config.toml#L27

maelle:
wait! It does break the URL https://deploy-preview-667--tidyverse-org.netlify.app/blog/2022/09/playing-on-the-same-team-as-your-dependency/

maelle:
I had stupidly updated the slug instead of the title :roll_eyes: :see_no_evil:

PR #679, opened 2024-01-29
Link to (google) privacy policy in footer

jennybc:
We need a link to the Google privacy policy on the tidyverse.org landing page and that is the overall goal of this PR. However, while I am here, I am doing the absolute *bare minimum* to begin the RStudio -> Posit transition that also needs to happen.

Currently "Privacy policy" has the wrong color (seems like it should be white) and I'm not sure why. And obviously its presence and placement is not aesthetically pleasing. This is not my wheelhouse, so I throw myself at your mercy @hadley to get this to a state you're willing to merge ASAP. We can tackle the broader and more proper RStudio --> Posit transition in a different PR. Where by "we" I mean "anyone but me" 😅


hadley:
I tweaked the css to make it look right. If you think it looks good, feel free to merge.

jennybc (commented):
Screenshot 2024-01-30 at 5 18 58 PM

LGTM!

PR #680, opened 2024-02-07
Fix broken link in blog post by using absolute URL

rkb965:
Hi, all! This is just a tiny link fix. I was appreciating a fab recent blog when I got a 404. Clicking on a {tidyclust} link sent me to https://www.tidyverse.org/blog/2023/11/tidymodels-errors-q4/tidyclust.tidymodels.org

I think (🤞) adding https:// fixes that


hadley:
Thanks!

PR #689, opened 2024-03-20
Add release post for webR 0.3.1

georgestagg:
To be posted after webR 0.3.0 hits npm and CDN.

* Indended to expand on some of the items in webR's `NEWS.md` file. The post is perhaps a little heavy on details aimed at JavaScript developers, but hopefully still interesting to the general audience.

* Some of the example output is hard-coded in the source, rather than generated by `knitr`, e.g. JS output and errors. There is some custom styling near the start of the post to tweak the visuals for this.

* Includes updates to the version of webR itself shipped on the website - used for interactive examples.

---

Updated links to hotfix 0.3.1 release.


content/blog/webr-0-3-1/index.md
```diff
@@ -0,0 +1,368 @@
...
+
+WebR's core functionality is to evaluate R code from a JavaScript environment. As such, it is imperative that this works well, even with large and complex scripts. The [webR app](https://webr.r-wasm.org/v0.3.1/) has been updated to better handle large R scripts, and scripts longer than 4096 characters should no longer cause strange issues in the R console.
+
+### Loading WebAssembly packages
+
+The package management functions provided by webR have been expanded and improved. Once enabled [^2], [`install.packages()`](https://rdrr.io/r/utils/install.packages.html), [`library()`](https://rdrr.io/r/base/library.html) and [`require()`](https://rdrr.io/r/base/library.html) are shimmed so that installing or loading R packages automatically downloads WebAssembly binaries from the [webR package repository](https://repo.r-wasm.org). Also, it is no longer required to run the [`library()`](https://rdrr.io/r/base/library.html) command a second time to subsequently load the package.
```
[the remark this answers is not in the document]
lionel-: > Also, it is no longer required to run the [`library()`](https://rdrr.io/r/base/library.html) command a second time to subsequently load the package.

🎉

content/blog/webr-0-3-1/index.md
```diff
@@ -0,0 +1,368 @@
...
+
+Improvements have been made to how webR raises R conditions as JavaScript exceptions. Exceptions now include the offending source R call in the error message text, better matching what is shown in a traditional R console.
+
+``` javascript
+await webR.evalR("sin('abc')");
+```
+
+<div class="output error">
+
+    Uncaught Error: Error in sin("abc"): non-numeric argument to mathematical function
```
lionel-, line 108: Unrelated to the post and not terribly important but could the call be backticked? Less consistent with base R but to be perfectly consistent we'd need a space between the call and semi-colon, I think it's better to follow the tidyverse style here.

content/blog/webr-0-3-1/index.md
```diff
@@ -0,0 +1,368 @@
...
+
+</div>
+
+Conditions raised when invoking function objects are now also re-thrown as JavaScript exceptions, rather than a generic `UnwindProtectException` error. Compare the error messages shown below from the previous and latest versions of webR to see the useful context added by this change.
+
+``` javascript
+// webR 0.2.2
+const do_calc = await webR.evalR(`function (n){ rnorm(n) }`)
```
lionel-, line 116: Tidyverse style nit:

```suggestion
const do_calc = await webR.evalR(`function (n) { rnorm(n) }`)
```

content/blog/webr-0-3-1/index.md
```diff
@@ -0,0 +1,368 @@
...
+<div class="output error">
+
+    Error in webr_hook_system(command) : 
+      The "system()" function is unsupported under Emscripten.
+
+</div>
+
+## Capturing HTML canvas graphics output
+
+The [`captureR()`](https://docs.r-wasm.org/webr/latest/evaluating.html#evaluating-r-code-and-capturing-output-with-capturer) function is designed to capture output generated when evaluating R code. With this release, plots drawn using webR's HTML canvas graphics device, [`webr::canvas()`](https://rdrr.io/pkg/webr/man/canvas.html), are also captured and returned by default.
```
lionel-, line 154: It might be interesting to mention the existing capture of errors and conditions, especially since the section just before is about errors.

content/blog/webr-0-3-1/index.md
```diff
@@ -0,0 +1,368 @@
...
+<div class="output">
+
+    [1] "raw"
+    [1] 0c
+
+</div>
+
+### R `data.frame`
+
+JavaScript objects of shape `{ x: [...], y: [...] }`, with data in a "long" column-based form, can now be used to construct R objects. In previous versions of webR, this object shape was reserved for future use. However, with this release webR now constructs an R `data.frame` by taking the source object's properties as column vectors. The resulting `data.frame` can then be manipulated from R in the usual way:
```
lionel-, line 230: The more I think about it the more I feel this would be better done with a helper ctor that would support both shapes of data frames and would check for data frame invariants on the JS side. Especially since this ctor downgrades to lists silently. It seems to me that the ease of use of this syntax is not worth the very real downsides on programming errors. We are compounding silent coercions and laziness of checking (if any) here.

If this is just for debugging then this should probably not be documented prominently. But also I feel like a short-name ctor would be just as good for interactivity.

lionel- (approved):
It is so great to see webR become mature software 😄

content/blog/webr-0-3-1/index.md
georgestagg: I agree, I don't think we need it to be entirely consistent with base R and I like the idea of using tidyverse style with backticks. I've made an issue to track this: https://github.com/r-wasm/webr/issues/397

content/blog/webr-0-3-1/index.md
georgestagg: As discussed in another forum, we'll keep this as-is for now, but address the issues in the next minor release of webR with the aim of stabilising the mapping between JS object shape and R object type.

My current plan is the following changes:

1) Reserve this JS syntax for constructing only `data.frame`s when using the generic `RObject()` constructor.
2) Remove the list fallback and throw an Error instead. R lists can still be constructed using e.g. `RList()`.
3) Add a small `RList` subclass, `RDataFrame`, for explicit `data.frame` construction.
3) Document the above, recommending for production code to use the explicit constructors rather than `RObject`.

We'll continue this discussion in a future PR on the webR repo.

content/blog/webr-0-3-1/index.md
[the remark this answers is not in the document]
georgestagg: I've simplified the sentence, removing the footnote. You're right, the link to the documentation makes it clear enough that this is configurable.

georgestagg:
Thanks again for the proof read and review!

PR #695, opened 2024-05-15
marquee release post

thomasp85:


content/blog/marquee-1-0-0/index.Rmd
```diff
@@ -0,0 +1,217 @@
...
+
+## Acknowledgements
+
+Marquee is the latest in a stream of advancements when it comes to text rendering and font support in R. It builds on top of my work with [systemfonts](https://systemfonts.r-lib.org/index.html), [textshaping](https://github.com/r-lib/textshaping), and [ragg](https://ragg.r-lib.org/index.html), but also pays great dept to Paul Murrell's work on adding a new, more low level API for text rendering to grid and the graphics engine. Lastly, Claus Wilke's work on [gridtext](https://wilkelab.org/gridtext/) and [ggtext](https://wilkelab.org/ggtext/) showed the power and need for rich text support in R and filled a gap until the technical foundation for marquee was build out.
```
hfrick, line 217:
```suggestion
Marquee is the latest in a stream of advancements when it comes to text rendering and font support in R. It builds on top of my work with [systemfonts](https://systemfonts.r-lib.org/index.html), [textshaping](https://github.com/r-lib/textshaping), and [ragg](https://ragg.r-lib.org/index.html), but also pays great dept to Paul Murrell's work on adding a new, more low level API for text rendering to grid and the graphics engine. Lastly, Claus Wilke's work on [gridtext](https://wilkelab.org/gridtext/) and [ggtext](https://wilkelab.org/ggtext/) showed the power and need for rich text support in R and filled a gap until the technical foundation for marquee was built out.
```

content/blog/marquee-1-0-0/index.Rmd
```diff
@@ -0,0 +1,217 @@
...
+```{r, include=FALSE}
+knitr::opts_chunk$set(
+  dpi = 300,
+  dev = "ragg_png"
+)
+```
+
+I am super excited to announce the initial release of [marquee](https://marquee.r-lib.org), a markdown parser and renderer for R graphics that allows native rich text formatting of text in graphics created with grid (which includes ggplot2 and lattice).
+
+The inception of this package goes all the way back to 2017
```
hadley, line 44:
```suggestion
The inception of this package goes all the way back to 2017:
```

content/blog/marquee-1-0-0/index.Rmd
```diff
@@ -0,0 +1,217 @@
...
+  geom_point(aes(x = mpg, y = disp)) + 
+  ggtitle(md_text) + 
+  theme(plot.title = element_marquee(size = 8, width = unit(16, "cm")))
+```
+
+## Styling
+
+As alluded to above, marquee comes with a styling API that is reminiscent of CSS but completely its own. In some sense it takes the "simplicity over power" approach from markdown and applies it to styling.
+
+In marquee, each element type (e.g. a code block) has its own style. This style can be incomplete in which case it inherits the remaining specifications from the parent element in the document. As an example, the `em` element has the following default style `style(italic = TRUE)`, that is, take whatever style is currently in effect but set the text to italic. This is basically how CSS works as well, but CSS allows so much more which clutters both the API and implementation if all you ever want to do is format rich text.
```
hadley, line 139:
```suggestion
In marquee, each element type (e.g. a code block) has its own style. This style can be incomplete in which case it inherits the remaining specifications from the parent element in the document. As an example, the `em` element has the following default style `style(italic = TRUE)`, that is, take whatever style is currently in effect but also make the text italic.
```

Don't think mention of CSS really adds anything here

content/blog/marquee-1-0-0/index.Rmd
```diff
@@ -0,0 +1,217 @@
...
+
+Apart from the direct inheritance of the marquee styling, it is also possible to use relative inheritance for numeric specifications (e.g. `lineheight = relative(2)` to double the current lineheight) or set sizes based on the current or root element font size (using `em()` and `rem()` respectively). Lastly, you can also mark a specification as "non-inheritable" using `skip_inherit()` in which case the inheritance moves up one level.
```
hadley, line 141: I don't get what this means.

content/blog/marquee-1-0-0/index.Rmd
```diff
@@ -0,0 +1,217 @@
...
+
+Marquee ships with a single style set (`classic_style()`) and you can make your own, either by modifying this, or building your own from scratch. If you choose the latter route make sure that the base element is a complete style - everything beyond that is optional.
```
hadley, line 143: I'd be somewhat tempted to just drop this in favour of pointing to the appropriate place in the marquee docs. It seems pretty advanced.

content/blog/marquee-1-0-0/index.Rmd
```diff
@@ -0,0 +1,217 @@
...
+md_plots <- 
+"# Plots
+In the plot below, the red dot (![](point)) shows the Mazda RX4
+
+![](plot)
+"
+
+grid::grid.draw(marquee_grob(md_plots))
+```
+
```
hadley, line 206: I think you should you should add something like:

> This means that you can now [include a ggplot2 plot in an annotation of another plot](https://www.youtube.com/watch?v=g3j9muCo4o0)!

hadley (approved):

content/blog/marquee-1-0-0/index.Rmd
thomasp85: if you set a value as `skip_inherit()` you say that any children should not inherit this value but instead the parent value. I'll try to come up with a better explanation

PR #728, opened 2025-03-18
add sparsity-tidymodels post

EmilHvitfeldt:
TODO: wait for tidymodels.org PRs to me merged and update links


topepo (commented):
Looks great. I made some text changes (mostly prompted by Grammarly). 

I think we should mention the process within recipes and the handoff from a recipe to a model. 

- _Internally_ we use a new sparse data format to store the data (when appropriate) as the recipe is being processed. 
- If the model can accept sparse objects, we convert the data from our new sparse format to a standard sparse matrix object.  

This leads me to a question: Will `bake()` ever output a data frame with sparse vectors? I want the post to say something like, “You’ll get the same data formats from `bake()` as you used to” but it is unclear. 

I went to `?bake()` to see if the help for the `composition` argument has changed. It hasn’t, but it probably needs to (emphasis is mine): 

> 	Either "tibble", "matrix", "data.frame", or '"dgCMatrix"“for the format of the processed data set. **Note that all computations during the baking process are done in a non-sparse format**. Also, note that this argument should be called after any selectors and the selectors should only resolve to numeric columns (otherwise an error is thrown).

I’ll put in an issue for that.

EmilHvitfeldt:
> This leads me to a question: Will bake() ever output a data frame with sparse vectors? I want the post to say something like, “You’ll get the same data formats from bake() as you used to” but it is unclear.

if you don't change the values of `sparse` and leave them at `"auto"` then `bake()` will produce what it used to before these changes. You have to opt-in by setting `sparse = "yes"`. which workflows can do you use if the model supports sparse data AND the data is sparse enough

PR #730, opened 2025-04-19
recipes 1.3.0 post

EmilHvitfeldt:


content/blog/recipes-1-3-0/index.Rmd
```diff
@@ -0,0 +1,183 @@
+---
+output: hugodown::hugo_document
+
+slug: recipes-1-3-0
+title: recipes 1.3.0
+date: 2025-04-22
+author: Emil Hvitfeldt
+description: >
+    This release brings multiple exciting features and streamlines many recipe steps.
```
hfrick, line 9: This feels a tad clickbait-y 😅

content/blog/recipes-1-3-0/index.Rmd
```diff
@@ -0,0 +1,183 @@
...
+
+We're thrilled to announce the release of [recipes](https://recipes.tidymodels.org/) 1.3.0. recipes lets you create a pipeable sequence of feature engineering steps.
+
+You can install it from CRAN with:
+
+```{r, eval = FALSE}
+install.packages("recipes")
+```
+
+This blog post will walk through some of the highlights of this release.
```
hfrick, line 43: I would put into words here what awaits the reader. That can help people decide if they want to read it, and when ("oh, I use `step_select()` all the time, I'm gonna read now to be informed"). Without that info, this sentence is very verbose to me and gives me very little info.

hfrick (approved):
I really like the change in how the contrasts are handled! That explicitness via an argument is so much better! 

(And the brownie picture looks delicious! Pictures for recipes blog posta are always some of the best! 🧁 )

content/blog/recipes-1-3-0/index.Rmd
```diff
@@ -0,0 +1,183 @@
...
+
+You can see a full list of changes in the [release notes](https://recipes.tidymodels.org/news/index.html#recipes-130).
+
+```{r setup, message=FALSE}
```
topepo, line 47: Please add a line above this saying something like "Let's first load the package:"

content/blog/recipes-1-3-0/index.Rmd
```diff
@@ -0,0 +1,183 @@
...
+library(recipes)
+```
+
+## `strings_as_factors`
+
+Recipes by default convert predictor strings to factors, and the option for that is located in `prep()`. This caused an issue when you wanted to set `strings_as_factors = FALSE` for a recipe that is used somewhere else like in a workflow.
+
+This is no longer an issue as we have moved the argument to `recipe()` itself. We are at the same time deprecating the use of `strings_as_factors` when used in `prep()`.
```
topepo, line 55:
```suggestion
This is no longer an issue as we have moved the argument to `recipe()` itself. We are at the same time deprecating the use of `strings_as_factors` when used in `prep()`. Here is an example:
```

content/blog/recipes-1-3-0/index.Rmd
```diff
@@ -0,0 +1,183 @@
...
+  step_dummy(all_nominal_predictors()) |>
+  prep() |>
+  bake(new_data = penguins)
+```
+
+The issue with this approach is that it pulls from `options()` when it needs it instead of storing the information. This means that if you put this recipe in production you will need to set the option in the production environment to match that of the training environment.
+
+```{r}
+#| echo: false
+options(contrasts = c(unordered = "contr.treatment", ordered = "contr.poly"))
```
topepo, line 129: Can you add a note on how to do this with a contrast in an external package?

topepo (commented):
👍

content/blog/recipes-1-3-0/index.Rmd
EmilHvitfeldt: added more information here https://github.com/tidyverse/tidyverse.org/pull/730/commits/be40d9d626ed8618b3ca9e47bed86274234befbd

content/blog/recipes-1-3-0/index.Rmd
EmilHvitfeldt: added more information here https://github.com/tidyverse/tidyverse.org/pull/730/commits/be40d9d626ed8618b3ca9e47bed86274234befbd

content/blog/recipes-1-3-0/index.Rmd
EmilHvitfeldt: done in https://github.com/tidyverse/tidyverse.org/pull/730/commits/18ee2a6c57084c1317996050fa3fe46cf716a931

content/blog/recipes-1-3-0/index.Rmd
EmilHvitfeldt: done in https://github.com/tidyverse/tidyverse.org/pull/730/commits/18ee2a6c57084c1317996050fa3fe46cf716a931

PR #735, opened 2025-05-26
Add post about ggplot2 migrating to S7

teunbrand:
This is a post we intend to release prior the new ggplot2 version. It'll be handy to point to for several revdepcheck failures that may need to adapt to ggplot2. It details some of the issues we expect package maintainers to encounter and possible solutions.
I'm tagging @thomasp85 for review and I'll send a reminder when we actually merge the S7 PR.


content/blog/ggplot2-4-0-0-s7/index.Rmd
```diff
@@ -0,0 +1,339 @@
...
+So if we want to change the stored `ink` colour, we can use:
+
+```{r}
+my_plot@ink <- "blue"
+```
+
+### Testing
+
+In S3, the recommended way to test for the class of an object is to use a testing function.
+For example `is.factor()`; and if such a testing function doesn't exist: use `inherits()`.
```
thomasp85, line 127:
```suggestion
An example is `is.factor()` but it may be that such a testing function doesn't exist:. In that case you can use `inherits()`.
```

content/blog/ggplot2-4-0-0-s7/index.Rmd
```diff
@@ -0,0 +1,339 @@
...
+  }
+)
+```
+
+The advice herein is thus to use `is_ggplot()`.
+
+## Generics and methods
+
+If you are new to object oriented programming in R, you might be unfamiliar what the terms 'generic' and 'methods' mean.
+They are a form of 'polymorphism', where we can use a single function, called the 'generic' function, with different implementations for different classes (where one such implementation is called a 'method').
```
thomasp85, line 210:
```suggestion
They are a form of 'polymorphism', that allow us to use a single function, called the 'generic' function, with different implementations for different classes (where one such implementation is called a 'method').
```

content/blog/ggplot2-4-0-0-s7/index.Rmd
```diff
@@ -0,0 +1,339 @@
...
+A well known generic is `print()`, which does different things for different classes. 
+For example `print(1:10)` prints the numeric vector to the console, but `print(my_plot)` opens a graphics device to render the plot.
```
thomasp85, line 212:
```suggestion
For example `print(1:10)` prints the numeric vector to the console, but `print(my_plot)` opens a graphics device and renders the plot.
```

PR #794, opened 2026-04-22
Fix disqus config for hugo 0.160

gaborcsardi:
I could not build the site with Hugo 0.160 or 0.155.  Is this needed? What hugo versions are people using?


gaborcsardi:
Aha, I see
```
      - name: Setup Hugo
        uses: peaceiris/actions-hugo@v3
        with:
          hugo-version: '0.71.1'
```
so we probably don't want a newer version?

